PR #16, opened 2023-08-21
issue: Support IP_BIND_ADDRESS_NO_PORT

iftahl:
1. Support IP_BIND_ADDRESS_NO_PORT in setsockopt, for both TCP and UDP.
2. Nginx initialization for UDP listen sockets that are not set to "reuseport".

## Change type
- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list

### DIFF
--- a/src/core/main.cpp
+++ b/src/core/main.cpp
@@ -152,6 +152,11 @@ static int free_libxlio_resources()
     }
     g_p_route_table_mgr = NULL;
 
+    if (g_bind_no_port) {
+        delete g_bind_no_port;
+    }
+    g_bind_no_port = NULL;
+
     if (g_p_rule_table_mgr) {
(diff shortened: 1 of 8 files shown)


igor-ivanov (commented):
Should not it be supported for UDP too?

iftahl:
> Should not it be supported for UDP too?

Yes, will be added later.

iftahl:
> Should not it be supported for UDP too?

You can look now.
UDP is much simpler than TCP because in connect() call we call os.call, so we can just use kernel's implementation for IP_BIND_NO_PORT. in TCP we don't call os.connect() so we needed to implement the DB for it.

PR #22, opened 2023-08-28
issue: 3580058 Add support of non-default hugepage sizes

pasis:
## Description
**TODO Update README with approved warning message on hugepage allocation failure.**

Generalize and improve xlio_allocator. Add support of non-default hugepage sizes.

##### What
Add support of non-default hugepage sizes.

##### Why ?
Utilize hugepages more efficiently and provide options to run XLIO with giant default hugepage size, especially in multi process environment.

## Change type
- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list


igor-ivanov (commented):
Probably there is a place to improve in current xlio logic related
MLX5_ environment variables usage such as `MLX4_DEVICE_FATAL_CLEANUP`, `MLX_QP_ALLOC_TYPE` etc.
XLIO explicitly set them instead of taking if it is set by user befor.

pasis:
> Probably there is a place to improve in current xlio logic related MLX5_ environment variables usage such as `MLX4_DEVICE_FATAL_CLEANUP`, `MLX_QP_ALLOC_TYPE` etc. XLIO explicitly set them instead of taking if it is set by user befor.

Changed: now XLIO doesn't override QP and CQ related variables. However, I would leave other variables hardcoded, because they control functional behavior, not just an optimization.

src/core/sock/tcp_seg_pool.cpp
```diff
@@ -41,23 +41,27 @@ extern global_stats_t g_global_stat_static;
 tcp_seg_pool *g_tcp_seg_pool = NULL;
 
 tcp_seg_pool::tcp_seg_pool(uint32_t size)
+    : m_allocator(ALLOC_TYPE_PREFER_HUGE)
 {
-    m_tcp_segs_array = new struct tcp_seg[size];
-    if (m_tcp_segs_array == NULL) {
+    tcp_seg *tcp_segs_array = (tcp_seg *)m_allocator.alloc(sizeof(tcp_seg) * size);
+    if (tcp_segs_array == nullptr) {
```
AlexanderGrissik, line 47: can be just !tcp_segs_array

src/core/sock/tcp_seg_pool.cpp
```diff
@@ -41,23 +41,27 @@ extern global_stats_t g_global_stat_static;
...
+        tcp_segs_array[i].next = &tcp_segs_array[i + 1];
     }
-    m_p_head = &m_tcp_segs_array[0];
+    m_p_head = &tcp_segs_array[0];
     g_global_stat_static.n_tcp_seg_pool_size = size;
 }
 
 tcp_seg_pool::~tcp_seg_pool()
 {
-    delete[] m_tcp_segs_array;
```
AlexanderGrissik, line 60: We don't deallocate because allocator will deallocate all its memory on exit? does it mean allocator will never deallocate allocated memory before exit?

pasis: xlio_allocator object deallocates memory in its destructor. So, once tcp_seg_pool object is destroyed, the relevant memory is freed.

`dealloc()` is a public method, so memory may be freed on demand before destructor if needed.

src/core/sock/tcp_seg_pool.cpp
pasis: done

src/core/util/hugepage_mgr.h
```diff
@@ -0,0 +1,129 @@
...
+#include <stddef.h>
+#include <stdint.h>
+
+#include <string>
+#include <vector>
+#include <unordered_map>
+
+#include "utils/lock_wrapper.h"
+
+typedef struct {
```
AlexanderGrissik, line 45: We don't need this construct with C++. struct hugepage_metadata {} is enough

src/core/util/hugepage_mgr.h
```diff
@@ -0,0 +1,129 @@
...
+
+    void read_sysfs();
+    size_t read_meminfo(const char *tag);
+    uint32_t read_file_uint(std::string path);
+
+    bool is_hugepage_optimal(size_t hugepage, size_t size);
+    bool is_hugepage_acceptable(size_t hugepage, size_t size);
+    size_t find_optimal_hugepage(size_t size);
+
+    inline size_t hugepage_unused_space(size_t hugepage, size_t size)
```
AlexanderGrissik, line 82: AFAIK inline keyword is no needed for in class implemented methods. the inline is implicit.

src/core/util/hugepage_mgr.h
```diff
@@ -0,0 +1,129 @@
...
+    {
+        const size_t hugepage_mask = hugepage - 1;
+        return ((size + hugepage_mask) & ~hugepage_mask) - size;
+    }
+
+    inline size_t hugepage_metric(size_t hugepage, size_t size)
+    {
+        // Percentage of the unused hugepage tail
+        return hugepage_unused_space(hugepage, size) / (hugepage / 100U);
```
AlexanderGrissik, line 91: Wouldn't it be better to first divide unused-space by hugepage, because hugepage / 100U will get unaccurate result because of integer division.

src/core/util/hugepage_mgr.h
```diff
@@ -0,0 +1,129 @@
...
+    size_t read_meminfo(const char *tag);
+    uint32_t read_file_uint(std::string path);
+
+    bool is_hugepage_optimal(size_t hugepage, size_t size);
+    bool is_hugepage_acceptable(size_t hugepage, size_t size);
+    size_t find_optimal_hugepage(size_t size);
+
+    inline size_t hugepage_unused_space(size_t hugepage, size_t size)
+    {
+        const size_t hugepage_mask = hugepage - 1;
```
AlexanderGrissik, line 84: Can you please add a comment that explains what this method does (why not just hugepage - size).

src/core/util/hugepage_mgr.cpp
```diff
@@ -0,0 +1,258 @@
...
+    vlog_printf(VLOG_INFO, "  Total: allocations=%u fails=%u\n", m_stats.allocations,
+                m_stats.fails);
+    vlog_printf(VLOG_INFO, "  Total: allocated=%zuMB requested=%zuMB unused_space=%zuMB\n",
+                m_stats.total_allocated / ONE_MB, m_stats.total_requested / ONE_MB,
+                m_stats.total_unused / ONE_MB);
+}
+
+void hugepage_mgr::read_sysfs()
+{
+    DIR *dir = opendir("/sys/kernel/mm/hugepages/");
```
AlexanderGrissik, line 207: Are these pathes the same for different OSs or we target only supported by XLIO?

src/core/util/hugepage_mgr.cpp
```diff
@@ -0,0 +1,258 @@
...
+    struct dirent *entry;
+
+    if (dir) {
+        while ((entry = readdir(dir))) {
+            if (strncmp(entry->d_name, "hugepages-", 10) == 0) {
+                std::string path = std::string("/sys/kernel/mm/hugepages/") + entry->d_name;
+                size_t key = atol(entry->d_name + 10) * 1024U;
+
+                m_hugepages[key].nr_hugepages_total = read_file_uint(path + "/nr_hugepages");
+                m_hugepages[key].nr_hugepages_free = read_file_uint(path + "/free_hugepages");
```
AlexanderGrissik, line 217: How do we track free hugepages if other applications allocate hugepages?

src/core/util/hugepage_mgr.cpp
```diff
@@ -0,0 +1,258 @@
...
+        ++m_stats.fails;
+    }
+    return ptr;
+}
+
+void hugepage_mgr::dealloc_hugepages(void *ptr, size_t size)
+{
+    int rc = munmap(ptr, size);
+    if (rc != 0) {
+        __log_info_dbg("munmap failed (errno=%d)", errno);
```
AlexanderGrissik, line 163: Isn't info/warn better for such case?

src/core/util/hugepage_mgr.cpp
```diff
@@ -0,0 +1,258 @@
...
+    }
+    // We don't track released hugepages. Usually they're used for preallocation
```
AlexanderGrissik, line 165: XLIO specific?

src/core/util/hugepage_mgr.cpp
```diff
@@ -0,0 +1,258 @@
...
+    size_t hugepage = find_optimal_hugepage(size);
+    size_t hugepage_mask = hugepage - 1;
+    size_t actual_size = size;
+    void *ptr = nullptr;
+
+    if (hugepage) {
+        int map_extra = 0;
+
+        if (hugepage != m_default_hugepage) {
+            map_extra = (int)log2(hugepage) << MAP_HUGE_SHIFT;
```
AlexanderGrissik, line 127: Can we rename it to something like map_non_default? map_extra sounds like we allocate extra size.

src/core/util/sys_vars.cpp
```diff
@@ -1734,35 +1739,20 @@ void mce_sys_var::get_env_params()
...
-        len = snprintf(mem_str, sizeof(mem_str), "%d", ALLOC_TYPE_HUGEPAGES);
-        if (likely((0 < len) && (len < (int)sizeof(mem_str)))) {
-            setenv(SYS_VAR_MEM_ALLOC_TYPE, mem_str,
-                   1); // Setenv to avoid core dump while valgrind is used.
+    if ((env_ptr = getenv(SYS_VAR_HUGEPAGE_LOG2)) != NULL) {
+        unsigned val = (unsigned)atoi(env_ptr);
+
+        // mmap() uses 6 bits for the hugepage size log2
+        if (val < 64U) {
+            hugepage_log2 = val;
```
AlexanderGrissik, line 1750: Do we want to check if it is a supported size? otherwise use default?

src/core/util/hugepage_mgr.h
pasis: Unused space is smaller than hugepage, so this would always be 0. We don't need a precise value for metric.

src/core/util/hugepage_mgr.h
pasis: done

src/core/util/hugepage_mgr.h
pasis: done.

`size` is the allocation request, it can be multiple hugepages. And the method returns unused space in the last hugepage because of alignment.

src/core/util/hugepage_mgr.cpp
```diff
@@ -0,0 +1,258 @@
...
+        }
+        actual_size = (size + hugepage_mask) & ~hugepage_mask;
+
+        __log_info_dbg("Allocating %zu bytes with hugepages %zu kB", actual_size, hugepage / 1024U);
+
+        ptr = mmap(NULL, actual_size, PROT_READ | PROT_WRITE,
```
AlexanderGrissik, line 133: is mmap always allocates complete hugepages or it can allocate memory from a page that is not completely used?

src/core/util/hugepage_mgr.cpp
```diff
@@ -0,0 +1,258 @@
...
+{
+    std::lock_guard<decltype(m_lock)> lock(m_lock);
+
+    m_hugepages.clear();
+    read_sysfs();
+}
+
+bool hugepage_mgr::is_hugepage_optimal(size_t hugepage, size_t size)
+{
+    return (size / hugepage) || hugepage_unused_space(hugepage, size) <= HUGEPAGE_UNUSED_OPTIMAL ||
```
AlexanderGrissik, line 70: Seems like this logic is relevant only if size < hugepage. but then isn't the samllest hugepage is best?

src/core/dev/buffer_pool.cpp
```diff
@@ -169,28 +165,27 @@ buffer_pool::buffer_pool(size_t buffer_count, size_t buf_size,
     memset(m_p_bpool_stat, 0, sizeof(*m_p_bpool_stat));
     xlio_stats_instance_create_bpool_block(m_p_bpool_stat);
 
-    if (buf_size == 0) {
+    if (buf_size == 0 || buffer_count == 0) {
```
AlexanderGrissik, line 168: think that else if can be just else

src/core/util/hugepage_mgr.cpp
pasis: We don't. There is a race and this is the reason why allocation may fail even if we predicated a success. Even running update() method before each allocation wouldn't avoid possible race with other processes

src/core/util/hugepage_mgr.cpp
pasis: I'd like to avoid noise. Usually, hugepages are freed at exit and system cleans up all the mappings at process termination.

src/core/util/hugepage_mgr.cpp
pasis: Yes, xlio_allocator is not a general purpose allocator. Without tracking released pages we don't have to lock this method and this also simplifies printing a report at XLIO termination (a future feature).

src/core/util/hugepage_mgr.h
pasis: removed keyword

src/core/util/sys_vars.cpp
pasis: Good idea, will implement

pasis: If requested hugepage (including non-zero `MCE_DEFAULT_HUGEPAGE_LOG2`) is not supported, XLIO will fallback to 0 (autodetection)

src/core/util/hugepage_mgr.cpp
pasis: mmap(MAP_HUGETLB) always allocates complete pages. So address is guaranteed to be aligned to the hugepage boundary and requested length is aligned if needed. However, we align length explicitly to (1) be able utilize the tail and (2) munmap() requires aligned length argument.

src/core/dev/allocator.cpp
```diff
@@ -30,392 +30,341 @@
...
-        free(m_data_block);
-        break;
-    default:
-        __log_info_err("Unknown memory allocation type %d", m_mem_alloc_type);
-        break;
-    }
-    __log_info_dbg("Done");
+/*static*/
+void xlio_allocator::initialize()
+{
```
AlexanderGrissik, line 79: empty method

src/core/util/hugepage_mgr.cpp
pasis: It's not an easy task to implement perfect optimal and acceptable criteria. We don't have enough information for this. After playing with 2 kinds of allocations simultaneously (buffer pools and buffer/tcp_seg metadata) I came with current naive approach.

Hugepages are checked in descending order, so the biggest optimal or the biggest acceptable will be used.

src/core/dev/buffer_pool.cpp
pasis: you're right, done

src/core/dev/allocator.cpp
pasis: Removed, will add it back if needed.

src/core/util/hugepage_mgr.cpp
AlexanderGrissik: Maybe we shouldn't bother with free pages at all. but selecting the next preferable huge-page size. lets discuss it.

AlexanderGrissik (approved):

PR #29, opened 2023-09-04
issue: 3580073 Add option_size for size parameters

pasis:
## Description
This is a part of hugepage optimization effort. PR is rebased on top of #22, only 2 commits are relevant.

Introduce option_size namespace with from_str() and to_str() methods to simplify and unify parameters which represent size. This approach follows other option-like implementations.

option_size::from_str() accepts dec, hex, oct formats and case insensitive suffixes B, K/KB, M/MB, G/GB. This is more convenient for big numbers.

##### What
Introduce option_size namespace with from_str() and to_str() methods.

##### Why ?
Simplify and unify parameters which represent size. Out of box user experience improvement.

## Change type
- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list

### DIFF
--- a/src/core/util/sys_vars.cpp
+++ b/src/core/util/sys_vars.cpp
@@ -312,7 +312,7 @@ OPTION_FROM_TO_STR_IMPL
 
 namespace option_tcp_ctl_thread {
 static option_t<mode_t> options[] = {
-    {CTL_THREAD_DISABLE, "Disable", {"disable", NULL, NULL}},
+    {CTL_THREAD_DISABLE, "Disabled", {"disable", "disabled", NULL}},
     {CTL_THREAD_DELEGATE_TCP_TIMERS, "Delegated TCP timers", {"delegate", NULL, NULL}},
     {CTL_THREAD_WITH_WAKEUP, "With Wakeup", {"with_wakeup", NULL, NULL}},
     {CTL_THREAD_NO_WAKEUP, "No Wakeup", {"no_wakeup", NULL, NULL}}};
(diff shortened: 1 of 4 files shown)


src/core/util/sys_vars.cpp
```diff
@@ -157,6 +157,75 @@ const char *to_str(xlio_spec_t level)
...
+                                                          ONE_MB,
+                                                      },
+                                                      {
+                                                          {"GB", "G"},
+                                                          ONE_GB,
+                                                      }};
+
+    for (auto &suffix : suffixes) {
+        for (auto sfx : suffix.vals) {
+            if (strcasecmp(sfx, endptr) == 0) {
```
AlexanderGrissik, line 203: Don't we want to support lower case?

src/core/util/sys_vars.cpp
```diff
@@ -1257,14 +1257,14 @@ void mce_sys_var::get_env_params()
     }
 
     if ((env_ptr = getenv(SYS_VAR_TX_BUF_SIZE)) != NULL) {
-        tx_buf_size = (uint32_t)atoi(env_ptr);
+        tx_buf_size = (uint32_t)option_size::from_str(env_ptr);
```
AlexanderGrissik, line 1260: why not use to_str in printing for these as well?

src/core/util/sys_vars.cpp
pasis: With strcasecmp() we support any combination of lower and upper case.

src/core/util/sys_vars.cpp
pasis: We can, willl update.

pasis: note that rx_buf_size is not displayed in the header. Updated only tx_buf_size and zc_tx_size

AlexanderGrissik (approved):

PR #33, opened 2023-09-11
issue: 3286324 Fixing stuck empty rx ring cleanup

AlexanderGrissik:
When the ring is destroyed and the RQ is empty the termination loop becomes infinte. It waits for the last_posted_wqe to be retrieved. but since it is empty the last posted wqe was already retrieved.

## Description
XLIO stuck infinitely in ring destruction

##### What
If the RQ is empty, ring destruction will loop inifintely.
See more details in commit message

##### Why ?
Fixing stuck XLIO

##### How ?
Try to fetch last posted wqe only if the RQ is not empty.

## Change type
- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list

### DIFF
--- a/src/core/dev/qp_mgr_eth_mlx5.h
+++ b/src/core/dev/qp_mgr_eth_mlx5.h
@@ -194,6 +194,8 @@ protected:
         return NULL;
     }
 
+    virtual bool is_rq_empty() const { return (m_mlx5_qp.rq.head == m_mlx5_qp.rq.tail); }
+
 private:
 #endif /* DEFINED_UTLS */
     inline int fill_wqe_send(xlio_ibv_send_wr *pswr);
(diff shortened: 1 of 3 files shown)


src/core/dev/qp_mgr_eth_mlx5.h
```diff
@@ -194,6 +194,8 @@ class qp_mgr_eth_mlx5 : public qp_mgr_eth {
         return NULL;
     }
 
+    virtual bool is_rq_empty() const { return (m_mlx5_qp.rq.head == m_mlx5_qp.rq.tail); }
```
pasis, line 197: [question] Is it safe to check without protection?

pasis (approved):

src/core/dev/qp_mgr_eth_mlx5.h
AlexanderGrissik: This method is called from safe place.

PR #48, opened 2023-10-12
issue: 3588192 Blocking API hanging with delegated TCP timers

AlexanderGrissik:
## Description
Several issues related Socketxtreme or delegated timers

##### What
1. Fixed wakeup waiting thread with Socketxtreme mode.
2. Fixed no TCP timers attempt in delegated mode for blocking calls.
3. Enhanced gtest CI to include More socketxtreme tests and delegated timers tests.

##### Why ?
Socketxterme/Delegate functionality

##### How ?
1. In case of socketxtreme wakeup method was skipped. wakeup must always be called regardless this is POSIX or socketxtreme API.
2. Added delegated timers attempt in rx_wait_helper method which is called for every blocking API. Thus, solving many scenarios that would hang with delegated timers and blocking calls.

## Change type
- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list

### DIFF
--- a/src/core/sock/sockinfo.h
+++ b/src/core/sock/sockinfo.h
@@ -412,12 +412,6 @@ protected:
     void remove_cqfd_from_sock_rx_epfd(ring *p_ring);
     int os_wait_sock_rx_epfd(epoll_event *ep_events, int maxevents);
     virtual bool try_un_offloading(); // un-offload the socket if possible
-    virtual inline void do_wakeup()
-    {
-        if (!is_socketxtreme()) {
-            wakeup_pipe::do_wakeup();
-        }
-    }
 
     bool is_shadow_socket_present() { return m_fd >= 0 && m_fd != m_rx_epfd; }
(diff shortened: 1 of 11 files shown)


src/core/sock/sockinfo.h
```diff
@@ -412,12 +412,6 @@ class sockinfo : public socket_fd_api,
     void remove_cqfd_from_sock_rx_epfd(ring *p_ring);
     int os_wait_sock_rx_epfd(epoll_event *ep_events, int maxevents);
     virtual bool try_un_offloading(); // un-offload the socket if possible
-    virtual inline void do_wakeup()
```
iftahl, line 415: so maybe after removing that, we can make it regular function instead of virtual.
we use it only in wakeup_pipe.

iftahl (approved):

AlexanderGrissik:
bot:retest

PR #62, opened 2023-11-15
Issue 3668182 productize lwip express data path

alexbriskin:
Productize an improved data path from express PoC

What
Get the LwiP-related changes

Why?
Reduce networking latency and CPU cache coherency.

## Change type
- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Checklist
- [ ] Comments have been inserted in hard-to-understand places

### DIFF
--- a/src/core/lwip/pbuf.c
+++ b/src/core/lwip/pbuf.c
@@ -162,16 +162,26 @@ u8_t pbuf_header(struct pbuf *p, s32_t header_size_increment)
         return 1;
     }
 
-    /* Check that we aren't going to move off the end of the pbuf */
-    if (header_size_increment < 0 && (-header_size_increment) > (s32_t)p->len) {
-        return 1;
-    }
+    if (header_size_increment >= 0) {
+        u32_t header_increment = (u32_t)header_size_increment;
+        /* set new payload pointer */
+        p->payload = (u8_t *)p->payload - header_increment;
+        /* modify pbuf length fields */
+        p->len += header_increment;
+        p->tot_len += header_increment;
+    } else {
+        u32_t header_decrement = (u32_t)(-header_size_increment);
+        /* Check that we aren't going to move off the end of the pbuf */
+        if (header_decrement > p->len) {
+            return 1;
+        }
 
-    /* set new payload pointer */
-    p->payload = (u8_t *)p->payload - header_size_increment;
-    /* modify pbuf length fields */
-    p->len += header_size_increment;
-    p->tot_len += header_size_increment;
+        /* set new payload pointer */
+        p->payload = (u8_t *)p->payload + header_decrement;
+        /* modify pbuf length fields */
+        p->len -= header_decrement;
+        p->tot_len -= header_decrement;
+    }
 
     LWIP_DEBUGF(PBUF_DEBUG | LWIP_DBG_TRACE,
                 ("pbuf_header: new %p (%" S32_F ")\n", (void *)p->payload, header_size_increment));
@@ -244,25 +254,6 @@ u8_t pbuf_free(struct pbuf *p)
         }
     }
     return count;
-}
-
-/**
- * Count number of pbufs in a chain
- *
- * @param p first pbuf of chain
- * @return the number of pbufs in a chain
- */
-
-u8_t pbuf_clen(struct pbuf *p)
-{
-    u8_t len;
-
-    len = 0;
-    while (p != NULL) {
-        ++len;
-        p = p->next;
-    }
-    return len;
 }
 
 /**
--- a/src/core/lwip/pbuf.c
+++ b/src/core/lwip/pbuf.c
@@ -257,6 +257,25 @@ u8_t pbuf_free(struct pbuf *p)
 }
 
 /**
+ * Count number of pbufs in a chain
+ *
+ * @param p first pbuf of chain
+ * @return the number of pbufs in a chain
+ */
+
+u8_t pbuf_clen(struct pbuf *p)
+{
+    u8_t len;
+
+    len = 0;
+    while (p != NULL) {
+        ++len;
+        p = p->next;
+    }
+    return len;
+}
+
+/**
  * Increment the reference count of the pbuf.
  *
  * @param p pbuf to increase reference counter of
--- a/src/core/lwip/pbuf.h
+++ b/src/core/lwip/pbuf.h
@@ -70,6 +70,7 @@ typedef struct {
         void *mdesc;
         int fd;
         u32_t mkey;
+        void *opaque;
     };
 } pbuf_desc;
 
--- a/src/core/lwip/pbuf.h
+++ b/src/core/lwip/pbuf.h
@@ -55,7 +55,6 @@ enum {
     PBUF_DESC_NONE = 0,
     PBUF_DESC_MDESC,
     PBUF_DESC_FD,
-    PBUF_DESC_MAP,
     PBUF_DESC_MKEY,
     PBUF_DESC_STRIDE,
     PBUF_DESC_TLS_RX,
@@ -65,6 +64,7 @@ enum {
 
 typedef struct {
     int attr;
+    u32_t express_mkey;
     union {
         void *map;
         void *mdesc;
--- a/src/core/lwip/pbuf.h
+++ b/src/core/lwip/pbuf.h
@@ -64,12 +64,11 @@ enum {
 
 typedef struct {
     int attr;
-    u32_t express_mkey;
+    u32_t mkey;
     union {
         void *map;
         void *mdesc;
         int fd;
-        u32_t mkey;
         void *opaque;
     };
 } pbuf_desc;
@@ -82,7 +81,7 @@ struct pbuf {
     void *payload;
 
     /** length of this buffer */
-    u16_t len;
+    u32_t len;
 
     u8_t gro;
 
@@ -129,7 +128,6 @@ void pbuf_realloc(struct pbuf *p, u32_t size);
 u8_t pbuf_header(struct pbuf *p, s32_t header_size);
 void pbuf_ref(struct pbuf *p);
 u8_t pbuf_free(struct pbuf *p);
-u8_t pbuf_clen(struct pbuf *p);
 void pbuf_cat(struct pbuf *head, struct pbuf *tail);
 
 void pbuf_split_64k(struct pbuf *p, struct pbuf **rest); // windows scale needs large pbuf
--- a/src/core/lwip/pbuf.h
+++ b/src/core/lwip/pbuf.h
@@ -60,6 +60,7 @@ enum {
     PBUF_DESC_STRIDE,
     PBUF_DESC_TLS_RX,
     PBUF_DESC_NVME_TX,
+    PBUF_DESC_EXPRESS,
 };
 
 typedef struct {
--- a/src/core/lwip/pbuf.h
+++ b/src/core/lwip/pbuf.h
@@ -128,6 +128,7 @@ void pbuf_realloc(struct pbuf *p, u32_t size);
 u8_t pbuf_header(struct pbuf *p, s32_t header_size);
 void pbuf_ref(struct pbuf *p);
 u8_t pbuf_free(struct pbuf *p);
+u8_t pbuf_clen(struct pbuf *p);
 void pbuf_cat(struct pbuf *head, struct pbuf *tail);
 
 void pbuf_split_64k(struct pbuf *p, struct pbuf **rest); // windows scale needs large pbuf
--- a/src/core/lwip/tcp_impl.h
+++ b/src/core/lwip/tcp_impl.h
@@ -296,7 +296,7 @@ struct tcp_seg {
 #define TF_SEG_OPTS_ZEROCOPY  (u8_t) TCP_WRITE_ZEROCOPY /* Use zerocopy send mode */
 
     u8_t tcp_flags; /* Cached TCP flags for outgoing segments */
-    u8_t bufs;
+    u8_t bufs; /* Number of buffers int the pbuf linked list */
 
     /* L2+L3+TCP header for zerocopy segments, it must have enough room for options
        This should have enough space for L2 (ETH+vLAN), L3 (IPv4/6), L4 (TCP)
--- a/src/core/lwip/tcp_impl.h
+++ b/src/core/lwip/tcp_impl.h
@@ -296,7 +296,6 @@ struct tcp_seg {
 #define TF_SEG_OPTS_ZEROCOPY  (u8_t) TCP_WRITE_ZEROCOPY /* Use zerocopy send mode */
 
     u8_t tcp_flags; /* Cached TCP flags for outgoing segments */
-    u8_t bufs; /* Number of buffers int the pbuf linked list */
 
     /* L2+L3+TCP header for zerocopy segments, it must have enough room for options
        This should have enough space for L2 (ETH+vLAN), L3 (IPv4/6), L4 (TCP)
--- a/src/core/lwip/tcp_impl.h
+++ b/src/core/lwip/tcp_impl.h
@@ -296,6 +296,7 @@ struct tcp_seg {
 #define TF_SEG_OPTS_ZEROCOPY  (u8_t) TCP_WRITE_ZEROCOPY /* Use zerocopy send mode */
 
     u8_t tcp_flags; /* Cached TCP flags for outgoing segments */
+    u8_t bufs; /* To avoid pbuf_clen() */
 
     /* L2+L3+TCP header for zerocopy segments, it must have enough room for options
        This should have enough space for L2 (ETH+vLAN), L3 (IPv4/6), L4 (TCP)
--- a/src/core/lwip/tcp_impl.h
+++ b/src/core/lwip/tcp_impl.h
@@ -57,7 +57,7 @@ void L3_level_tcp_input(struct pbuf *p, struct tcp_pcb *pcb);
 
 /* Used within the TCP code only: */
 struct tcp_pcb *tcp_alloc(u8_t prio);
-struct pbuf *tcp_tx_pbuf_alloc(struct tcp_pcb *pcb, u16_t length, pbuf_type type, pbuf_desc *desc,
+struct pbuf *tcp_tx_pbuf_alloc(struct tcp_pcb *pcb, u32_t length, pbuf_type type, pbuf_desc *desc,
                                struct pbuf *p_buff);
 void tcp_tx_preallocted_buffers_free(struct tcp_pcb *pcb);
 void tcp_tx_pbuf_free(struct tcp_pcb *pcb, struct pbuf *pbuf);
@@ -296,7 +296,7 @@ struct tcp_seg {
 #define TF_SEG_OPTS_ZEROCOPY  (u8_t) TCP_WRITE_ZEROCOPY /* Use zerocopy send mode */
 
     u8_t tcp_flags; /* Cached TCP flags for outgoing segments */
-    u8_t bufs; /* To avoid pbuf_clen() */
+    u8_t bufs;
 
     /* L2+L3+TCP header for zerocopy segments, it must have enough room for options
        This should have enough space for L2 (ETH+vLAN), L3 (IPv4/6), L4 (TCP)
--- a/src/core/lwip/tcp_in.c
+++ b/src/core/lwip/tcp_in.c
@@ -176,6 +176,7 @@ void L3_level_tcp_input(struct pbuf *p, struct tcp_pcb *pcb)
             in_data.inseg.seqno = in_data.seqno;
             in_data.inseg.flags = 0;
             in_data.inseg.tcp_flags = in_data.flags;
+            in_data.inseg.bufs = 0;
 
             in_data.recv_data = NULL;
             in_data.recv_flags = 0;
--- a/src/core/lwip/tcp_in.c
+++ b/src/core/lwip/tcp_in.c
@@ -176,7 +176,6 @@ void L3_level_tcp_input(struct pbuf *p, struct tcp_pcb *pcb)
             in_data.inseg.seqno = in_data.seqno;
             in_data.inseg.flags = 0;
             in_data.inseg.tcp_flags = in_data.flags;
-            in_data.inseg.bufs = 0;
 
             in_data.recv_data = NULL;
             in_data.recv_flags = 0;
@@ -941,7 +940,6 @@ static u32_t tcp_shrink_segment(struct tcp_pcb *pcb, struct tcp_seg *seg, u32_t 
                 ("tcp_shrink: count: %-5d unsent %s\n", count, _dump_seg(pcb->unsent)));
 #endif /* TCP_TSO_DEBUG */
 
-    seg->bufs -= count;
     return count;
 }
 
@@ -988,7 +986,6 @@ static u32_t tcp_shrink_zc_segment(struct tcp_pcb *pcb, struct tcp_seg *seg, u32
     }
     seg->tcphdr->seqno = htonl(seg->seqno);
 
-    seg->bufs -= count;
     return count;
 }
 
@@ -1210,12 +1207,14 @@ static void tcp_receive(struct tcp_pcb *pcb, tcp_in_data *in_data)
                 pcb->unacked = pcb->unacked->next;
                 LWIP_DEBUGF(TCP_QLEN_DEBUG,
                             ("tcp_receive: queuelen %" U32_F " ... ", (u32_t)pcb->snd_queuelen));
+                LWIP_ASSERT("pcb->snd_queuelen >= pbuf_clen(next->p)",
+                            (pcb->snd_queuelen >= pbuf_clen(next->p)));
                 /* Prevent ACK for FIN to generate a sent event */
                 if ((pcb->acked != 0) && ((next->tcp_flags & TCP_FIN) != 0)) {
                     pcb->acked--;
                 }
 
-                pcb->snd_queuelen -= next->bufs;
+                pcb->snd_queuelen -= pbuf_clen(next->p);
                 tcp_tx_seg_free(pcb, next);
                 LWIP_DEBUGF(TCP_QLEN_DEBUG,
                             ("%" U32_F " (after freeing unacked)\n", (u32_t)pcb->snd_queuelen));
@@ -1261,11 +1260,13 @@ static void tcp_receive(struct tcp_pcb *pcb, tcp_in_data *in_data)
             pcb->unsent = pcb->unsent->next;
             LWIP_DEBUGF(TCP_QLEN_DEBUG,
                         ("tcp_receive: queuelen %" U32_F " ... ", (u32_t)pcb->snd_queuelen));
+            LWIP_ASSERT("pcb->snd_queuelen >= pbuf_clen(next->p)",
+                        (pcb->snd_queuelen >= pbuf_clen(next->p)));
             /* Prevent ACK for FIN to generate a sent event */
             if ((pcb->acked != 0) && ((next->tcp_flags & TCP_FIN) != 0)) {
                 pcb->acked--;
             }
-            pcb->snd_queuelen -= next->bufs;
+            pcb->snd_queuelen -= pbuf_clen(next->p);
             tcp_tx_seg_free(pcb, next);
             LWIP_DEBUGF(TCP_QLEN_DEBUG,
                         ("%" U16_F " (after freeing unsent)\n", (u32_t)pcb->snd_queuelen));
--- a/src/core/lwip/tcp_in.c
+++ b/src/core/lwip/tcp_in.c
@@ -941,6 +941,7 @@ static u32_t tcp_shrink_segment(struct tcp_pcb *pcb, struct tcp_seg *seg, u32_t 
                 ("tcp_shrink: count: %-5d unsent %s\n", count, _dump_seg(pcb->unsent)));
 #endif /* TCP_TSO_DEBUG */
 
+    seg->bufs -= count;
     return count;
 }
 
@@ -987,6 +988,7 @@ static u32_t tcp_shrink_zc_segment(struct tcp_pcb *pcb, struct tcp_seg *seg, u32
     }
     seg->tcphdr->seqno = htonl(seg->seqno);
 
+    seg->bufs -= count;
     return count;
 }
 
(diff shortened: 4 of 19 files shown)


alexbriskin:
bot:retest

alexbriskin:
bot:retest

src/core/sock/sockinfo_ulp.cpp
```diff
@@ -752,32 +754,25 @@ ssize_t sockinfo_tcp_ops_tls::tx(xlio_tx_call_attr_t &tx_arg)
...
+            size_t tosend = std::min<size_t>(p_iov[i].iov_len - pos, TLS_RECORD_MAX);
 
-            /*
-             * XXX This approach can lead to issue with epoll()
-             * since such a socket will always be ready for write
-             */
-            if (!block_this_run && sndbuf < TLS_RECORD_SMALLEST &&
-                (sndbuf < m_tls_rec_overhead || (sndbuf - m_tls_rec_overhead) < tosend)) {
-                /* We don't want to create too small TLS records when we do partial write. */
+            if (m_p_sock->sndbuf_available() == 0U) {
```
alexbriskin, line 766:
```suggestion
            if (m_p_sock->sndbuf_available() == 0U && !block_this_run) {
```

alexbriskin:
Solved conflicts and rebased on top of jan24 branch

src/core/lwip/pbuf.h
```diff
@@ -55,20 +55,22 @@ enum {
...
 };
 
 typedef struct {
     int attr;
+    u32_t express_mkey;
     union {
         void *map;
         void *mdesc;
         int fd;
         u32_t mkey;
```
pasis, line 72: please move mkey to express_mkey

src/core/lwip/tcp_out.c
```diff
@@ -738,6 +742,122 @@ err_t tcp_write(struct tcp_pcb *pcb, const void *arg, u32_t len, u16_t apiflags,
...
+
+    if (len < pcb->mss) {
+        const int byte_queued = pcb->snd_nxt - pcb->lastack;
+        pcb->snd_sml_add = (pcb->unacked ? pcb->unacked->len : 0) + byte_queued;
+    }
+
+    /*
+     * TCP segmentation is done in three phases with increasing complexity:
+     *
+     * 1. Copy data directly into an oversized pbuf.
```
pasis, line 763: Please update comment to represent current functionality.

src/core/lwip/pbuf.h
alexbriskin: Done

src/core/lwip/tcp_out.c
alexbriskin: Done

src/core/lwip/err.h
```diff
@@ -48,6 +48,7 @@ typedef s8_t err_t;
 
 /* Definitions for error constants. */
 
+#define ERR_UNKNOWN    1 /* Unknown error.       */
```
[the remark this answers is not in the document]
alexbriskin: Done

src/core/sock/sockinfo_tcp.cpp
```diff
@@ -6042,3 +6029,192 @@ inline bool sockinfo_tcp::handle_bind_no_port(int &bind_ret, in_port_t in_port,
...
+    mdesc.opaque = opaque_op;
+
+    int bytes_written = 0;
+
+    lock_tcp_con();
+
+    for (unsigned i = 0; i < iov_len; ++i) {
+        err = tcp_write_express(&m_pcb, iov[i].iov_base, iov[i].iov_len, &mdesc);
+        if (err != ERR_OK) {
+            return -1;
```
pasis, line 6076: We need either:
1. Move socket to an error state, so user may not work with it
2. Or revert all the tcp_write_express() segments, so we don't corrupt TCP stream with send() retry.

src/core/sock/sockinfo_tcp.cpp
```diff
@@ -6042,3 +6029,192 @@ inline bool sockinfo_tcp::handle_bind_no_port(int &bind_ret, in_port_t in_port,
...
+        }
+        bytes_written += iov[i].iov_len;
+    }
+
+    if (!(flags & XLIO_EXPRESS_MSG_MORE)) {
+        err = tcp_output(&m_pcb);
+        if (err != ERR_OK) {
+            return -1;
```
pasis, line 6084: The same as above, if we create TCP segments with tcp_write() and don't support partial write, we need either do revert to the segments or move socket to error state.

Review what errors tcp_output() can return and if it returns recoverable error - return bytes_written, because we already created all the segments. This may lead to a deadlock in a corner case thoug...

alexbriskin:
bot:retest

src/core/sock/sockinfo_tcp.cpp
```diff
@@ -6042,3 +6024,211 @@ inline bool sockinfo_tcp::handle_bind_no_port(int &bind_ret, in_port_t in_port,
...
+            /* The only error in tcp_write_express is a memory error */
+            return tcp_tx_handle_errno_and_unlock(ENOMEM);
+        }
+        bytes_written += iov[i].iov_len;
+    }
+
+    if (!(flags & XLIO_EXPRESS_MSG_MORE)) {
+        err = tcp_output(&m_pcb);
+        if (err != ERR_OK) {
+            /* The error very likely to be recoverable */
```
pasis, line 6097: [Note for future] Need to handle error and revert all added tcp segments - we don't expect memory allocation errors, so this can be slow.

src/core/lwip/tcp_impl.h
```diff
@@ -296,6 +296,7 @@ struct tcp_seg {
 #define TF_SEG_OPTS_ZEROCOPY  (u8_t) TCP_WRITE_ZEROCOPY /* Use zerocopy send mode */
 
     u8_t tcp_flags; /* Cached TCP flags for outgoing segments */
+    u8_t bufs;
```
AlexanderGrissik, line 299: Please add a comment to describe what this field is for.

src/core/lwip/pbuf.h
```diff
@@ -80,7 +82,7 @@ struct pbuf {
     void *payload;
 
     /** length of this buffer */
-    u16_t len;
+    u32_t len;
```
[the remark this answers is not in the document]
AlexanderGrissik: The above should work with len of u16_t because the + operator takes the stronger type.
Is there another reason to change it to u32_t?

src/core/lwip/pbuf.c
```diff
@@ -162,16 +162,26 @@ u8_t pbuf_header(struct pbuf *p, s32_t header_size_increment)
...
-    }
+    if (header_size_increment >= 0) {
+        u32_t header_increment = (u32_t)header_size_increment;
+        /* set new payload pointer */
+        p->payload = (u8_t *)p->payload - header_increment;
+        /* modify pbuf length fields */
+        p->len += header_increment;
+        p->tot_len += header_increment;
+    } else {
+        u32_t header_decrement = (u32_t)(-header_size_increment);
```
AlexanderGrissik, line 173: What will happen when header_size_increment>0?

src/core/lwip/pbuf.c
```diff
@@ -162,16 +162,26 @@ u8_t pbuf_header(struct pbuf *p, s32_t header_size_increment)
         return 1;
     }
 
-    /* Check that we aren't going to move off the end of the pbuf */
-    if (header_size_increment < 0 && (-header_size_increment) > (s32_t)p->len) {
-        return 1;
-    }
+    if (header_size_increment >= 0) {
```
AlexanderGrissik, line 165: Not sure this refactoring is good in terms of performance. More code lines to fit in cache.

src/core/lwip/tcp_in.c
```diff
@@ -176,6 +176,7 @@ void L3_level_tcp_input(struct pbuf *p, struct tcp_pcb *pcb)
             in_data.inseg.seqno = in_data.seqno;
             in_data.inseg.flags = 0;
             in_data.inseg.tcp_flags = in_data.flags;
+            in_data.inseg.bufs = 0;
```
AlexanderGrissik, line 179: Do we set it to 0 because this is RX?

src/core/lwip/tcp_in.c
```diff
@@ -1375,7 +1376,7 @@ static void tcp_receive(struct tcp_pcb *pcb, tcp_in_data *in_data)
                adjust the ->data pointer in the seg and the segment
                length.*/
 
-            off = pcb->rcv_nxt - in_data->seqno;
+            u32_t off = pcb->rcv_nxt - in_data->seqno;
```
AlexanderGrissik, line 1379: Are we sure that pcb->rcv_nxt = in_data->seqno. What if malicious app sends invalid packet with bigger seqno?

src/core/proto/dst_entry_tcp.cpp
```diff
@@ -222,7 +222,9 @@ ssize_t dst_entry_tcp::fast_send(const iovec *p_iov, const ssize_t sz_iov, xlio_
             m_sge[i].addr = (uintptr_t)p_tcp_iov[i].iovec.iov_base;
             m_sge[i].length = p_tcp_iov[i].iovec.iov_len;
             if (is_zerocopy) {
-                if (PBUF_DESC_MKEY == p_tcp_iov[i].p_desc->lwip_pbuf.pbuf.desc.attr) {
+                if (PBUF_DESC_EXPRESS == p_tcp_iov[i].p_desc->lwip_pbuf.pbuf.desc.attr) {
```
AlexanderGrissik, line 225: Can we refactor to more readable
`auto desc_attr = p_tcp_iov[i].p_desc->lwip_pbuf.pbuf.desc.attr`
if (PBUF_DESC_EXPRESS == desc_attr || PBUF_DESC_MKEY == desc_attr)
 ....
else if (PBUF_DESC_MDESC == desc_attr || ...)
  ....

src/core/sock/sockinfo_ulp.cpp
```diff
@@ -752,32 +754,25 @@ ssize_t sockinfo_tcp_ops_tls::tx(xlio_tx_call_attr_t &tx_arg)
...
+    uint8_t *iv = is_tx_tls13() ? nullptr : m_tls_info_tx.iv;
+    mem_desc *zc_owner = is_zerocopy ? reinterpret_cast<mem_desc *>(tx_arg.priv.mdesc) : nullptr;
     for (ssize_t i = 0; i < tx_arg.attr.sz_iov; ++i) {
         pos = 0;
         while (pos < p_iov[i].iov_len) {
             tls_record *rec;
             ssize_t ret2;
-            size_t sndbuf = m_p_sock->sndbuf_available();
-            size_t tosend = p_iov[i].iov_len - pos;
+            size_t tosend = std::min<size_t>(p_iov[i].iov_len - pos, TLS_RECORD_MAX);
```
AlexanderGrissik, line 764: Question: Can TLS record have 16K payload or 16K should include the TLS header/trailer?

pasis: according to RFC 8446 sections 5.1 and 5.2, this is limitation of the plaintext payload (before adding inner record type in TLS1.3). TLS1.2 describes limits similarly.

src/core/lwip/tcp_in.c
pasis: I believe there is a check for this somewhere at the top of the function.

src/core/lwip/tcp_out.c
```diff
@@ -539,14 +554,10 @@ err_t tcp_write(struct tcp_pcb *pcb, const void *arg, u32_t len, u16_t apiflags,
...
-            if (is_zerocopy) {
-                concat_p->payload = (u8_t *)arg + pos;
-            } else {
-                memcpy(concat_p->payload, (u8_t *)arg + pos, seglen);
-            }
+            memcpy(concat_p->payload, (u8_t *)arg + pos, seglen);
 
             pos += seglen;
-            queuelen += pbuf_clen(concat_p);
+            queuelen++;
```
AlexanderGrissik, line 560: shouldn't be queuelen += concat_p->bufs ?

src/core/sock/sockinfo_tcp.cpp
```diff
@@ -894,80 +889,163 @@ ssize_t sockinfo_tcp::tx(xlio_tx_call_attr_t &tx_arg)
...
+    }
+
+    bool is_dummy = IS_DUMMY_PACKET(flags);
+    bool is_blocking = BLOCK_THIS_RUN(m_b_blocking, flags);
+    bool is_packet_zerocopy = (flags & MSG_ZEROCOPY) && ((m_b_zc) || (tx_arg.opcode == TX_FILE));
+    if (unlikely(is_dummy) || unlikely(!is_packet_zerocopy) || unlikely(is_blocking)) {
+        return tcp_tx_slow_path(tx_arg);
+    }
+
+    bool is_send_zerocopy = tx_arg.opcode != TX_FILE;
```
AlexanderGrissik, line 951: we need to rename this var to is_user_zc or something like that. becuase file is also zero copy just not from user.

src/core/sock/sockinfo_tcp.cpp
```diff
@@ -1048,43 +1118,17 @@ ssize_t sockinfo_tcp::tcp_tx(xlio_tx_call_attr_t &tx_arg)
...
                     si_tcp_logdbg("TX on disconnected socket");
-                    errno = ECONNRESET;
-                    goto err;
+                    return tcp_tx_handle_errno_and_unlock(ECONNRESET);
                 }
                 // force out TCP data before going on wait()
                 tcp_output(&m_pcb);
 
                 /* Set return values for nonblocking socket and finish processing */
-                if (!block_this_run) {
```
AlexanderGrissik, line 1058: Shouldn't we call here tcp_tx_handle_sndbuf_unavailable?

src/core/sock/sockinfo_tcp.cpp
```diff
@@ -1095,63 +1139,48 @@ ssize_t sockinfo_tcp::tcp_tx(xlio_tx_call_attr_t &tx_arg)
...
-                // transmit thread.
-                poll_count = 0;
+                    // AlexV:Avoid from going to sleep, for the blocked socket of course, since
+                    // progress engine may consume an arrived credit and it will not wakeup the
+                    // transmit thread.
+                    poll_count = 0;
 
-                goto retry_write;
-            }
+                    continue;
```
AlexanderGrissik, line 1180: what reason do we have to loop here?

src/core/sock/sockinfo_tcp.cpp
```diff
@@ -1095,63 +1139,48 @@ ssize_t sockinfo_tcp::tcp_tx(xlio_tx_call_attr_t &tx_arg)
...
-                    si_tcp_logpanic("tcp_write return: %d", err);
-                    BULLSEYE_EXCLUDE_BLOCK_END
+
+                if (apiflags & XLIO_TX_PACKET_ZEROCOPY) {
+                    err = tcp_write_express(&m_pcb, tx_ptr, tx_size, &tx_arg.priv);
+                } else {
+                    err = tcp_write(&m_pcb, tx_ptr, tx_size, apiflags, &tx_arg.priv);
                 }
-                /* Set return values for nonblocking socket and finish processing */
-                if (!block_this_run) {
```
AlexanderGrissik, line 1137: What about non blocking sockets?

src/core/sock/sockinfo_tcp.cpp
```diff
@@ -4124,21 +4116,11 @@ void sockinfo_tcp::fit_rcv_wnd(bool force_fit)
 
 void sockinfo_tcp::fit_snd_bufs(unsigned int new_max_snd_buff)
 {
-    uint32_t sent_buffs_num = 0;
+    m_pcb.snd_buf += (new_max_snd_buff - m_pcb.max_snd_buff);
+    m_pcb.max_snd_buff = new_max_snd_buff;
 
-    sent_buffs_num = m_pcb.max_snd_buff - m_pcb.snd_buf;
```
AlexanderGrissik, line 4129: Is m_pcb.snd_buf always >0

src/core/sock/sockinfo_tcp.cpp
```diff
@@ -894,80 +889,163 @@ ssize_t sockinfo_tcp::tx(xlio_tx_call_attr_t &tx_arg)
...
-    return false;
+#define TCP_WND_UNAVALABLE(pcb, total_iov_len) false
 #endif
+
+static inline bool is_invalid_iovec(const iovec *iov, size_t sz_iov)
+{
+    return iov == nullptr || sz_iov == 0;
 }
 
 ssize_t sockinfo_tcp::tcp_tx(xlio_tx_call_attr_t &tx_arg)
```
AlexanderGrissik, line 911: We have now 3 pathes. tcp_tx tcp_tx_slow_path and tcp_tx_express. When tcp_tx is ZC. don't we want to utilize tcp_tx_express for tcp_tx? why must we have 3 pathes?

alexbriskin:
bot:retest

src/core/sock/sockinfo_ulp.cpp
```diff
@@ -752,32 +754,25 @@ ssize_t sockinfo_tcp_ops_tls::tx(xlio_tx_call_attr_t &tx_arg)
...
 
-            /*
-             * XXX This approach can lead to issue with epoll()
-             * since such a socket will always be ready for write
-             */
-            if (!block_this_run && sndbuf < TLS_RECORD_SMALLEST &&
-                (sndbuf < m_tls_rec_overhead || (sndbuf - m_tls_rec_overhead) < tosend)) {
-                /* We don't want to create too small TLS records when we do partial write. */
+            if (m_p_sock->sndbuf_available() == 0U && !block_this_run) {
```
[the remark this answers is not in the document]
AlexanderGrissik: [For future] Think where it can be described.

src/core/sock/sockinfo_tcp.cpp
pasis: not in context of current PR.

src/core/sock/sockinfo_tcp.cpp
```diff
@@ -895,80 +890,168 @@ ssize_t sockinfo_tcp::tx(xlio_tx_call_attr_t &tx_arg)
...
+    int errno_tmp = errno;
+    int poll_count = 0;
+    uint16_t apiflags = 0;
+    err_t err;
+    bool is_send_zerocopy = false;
+    void *tx_ptr = NULL;
+    struct xlio_pd_key *pd_key_array = NULL;
 
     if (m_sysvar_rx_poll_on_tx_tcp) {
         rx_wait_helper(poll_count, false);
```
AlexanderGrissik, line 1057: This s already done in tcp_tx

src/core/sock/sockinfo_tcp.cpp
```diff
@@ -1003,30 +1086,23 @@ ssize_t sockinfo_tcp::tcp_tx(xlio_tx_call_attr_t &tx_arg)
...
-                        [](size_t sum, const iovec &curr) { return sum + curr.iov_len; });
     lock_tcp_con();
 
-    if (cannot_do_requested_dummy_send(m_pcb, tx_arg) ||
-        cannot_do_requested_partial_write(m_pcb, tx_arg, m_b_blocking, total_iov_len) ||
-        tcp_wnd_unavalable(m_pcb, total_iov_len)) {
-        unlock_tcp_con();
-        errno = EAGAIN;
-        return -1;
+    if (cannot_do_requested_dummy_send(m_pcb, tx_arg) || TCP_WND_UNAVALABLE(m_pcb, total_iov_len)) {
```
AlexanderGrissik, line 1091: total_iov_len is undefined in case TCP_WND_UNAVALABLE defined.
Do we really need TCP_WND_UNAVALABLE? Do we ever use it?

src/core/sock/sockinfo_tcp.cpp
```diff
@@ -1096,63 +1150,57 @@ ssize_t sockinfo_tcp::tcp_tx(xlio_tx_call_attr_t &tx_arg)
...
+                    return tcp_tx_handle_partial_send_and_unlock(total_tx, EINTR, is_dummy,
+                                                                 is_send_zerocopy, errno_tmp);
                 }
-                if (unlikely(err != ERR_MEM)) {
-                    // we should not get here...
-                    BULLSEYE_EXCLUDE_BLOCK_START
-                    si_tcp_logpanic("tcp_write return: %d", err);
-                    BULLSEYE_EXCLUDE_BLOCK_END
+
+                if (apiflags & XLIO_TX_PACKET_ZEROCOPY) {
```
AlexanderGrissik, line 1165: Maybe more c++ style:
err = (apiflags & XLIO_TX_PACKET_ZEROCOPY)
   ? tcp_write_express(&m_pcb, tx_ptr, tx_size, &tx_arg.priv)
   : tcp_write(&m_pcb, tx_ptr, tx_size, apiflags, &tx_arg.priv)

src/core/sock/sockinfo_tcp.cpp
```diff
@@ -6046,3 +6047,204 @@ inline bool sockinfo_tcp::handle_bind_no_port(int &bind_ret, in_port_t in_port,
...
+            si_tcp_logdbg("TX on disconnected socket");
+            errno = EPIPE;
+        }
+        return -1;
+    }
+
+    err_t err;
+    pbuf_desc mdesc;
+
+    if (unlikely(!is_rts())) {
```
AlexanderGrissik, line 6074: Double check

src/core/sock/sockinfo_tcp.cpp
```diff
@@ -6046,3 +6047,204 @@ inline bool sockinfo_tcp::handle_bind_no_port(int &bind_ret, in_port_t in_port,
 
     return CONTINUE_WITH_BIND;
 }
+
+int sockinfo_tcp::tcp_tx_express(const struct iovec *iov, unsigned iov_len, uint32_t mkey,
+                                 xlio_express_flags flags, void *opaque_op)
+{
+    if (unlikely(!is_rts())) {
```
AlexanderGrissik, line 6054: Why not use is_connected_and_ready_to_send() with param if to do wait_rx. We better have a single point of truce.

src/core/sock/sockinfo_tcp.cpp
AlexanderGrissik: I guess we need to check here if the error is ERR_MEM and then to revert posted segments. Because we must not allow partial write and in case of out of mem we have no choice.
I think it can be as separate PR. But we must handle this case somehow.

src/core/lwip/tcp_in.c
AlexanderGrissik: Shouldn't it be 1 in such case?
does bufs say how much chained including the current one or excluding? Worth to add this info to the comment of this var.

src/core/sock/sockinfo_ulp.cpp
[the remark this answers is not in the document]
alexbriskin: Let's add the comment in the tcp_pcb struct. *_express functions require available sndbuf (comment) + description for difference between the fast/slow/express use cases.

src/core/sock/sockinfo_tcp.cpp
alexbriskin: Adding appropriate comments

src/core/sock/sockinfo_tcp.cpp
alexbriskin: If it's redundant, I'll remove it

src/core/sock/sockinfo_tcp.cpp
alexbriskin: Let's put the total_iov_len in the same place as the DEFINED_TCP_TX_WND_AVAILABILITY exists.

src/core/sock/sockinfo_tcp.cpp
alexbriskin: Replace with ternary expression.

src/core/sock/sockinfo_tcp.cpp
alexbriskin: Got it

src/core/sock/sockinfo_tcp.cpp
alexbriskin: I'll explore it

AlexanderGrissik:
seg->nbufs should be checked for correctness with each tcp_create_segment and currently not sure it will be handled well in all cases. For this PR we better remove this change and do it as sperate PR putting focus into this.

src/core/sock/sockinfo_tcp.cpp
AlexanderGrissik: I think for this stage option 1 is good.

AlexanderGrissik:
See reply for 'Move socket to an error state, so user may not work with it

src/core/lwip/tcp_impl.h
```diff
@@ -296,6 +296,7 @@ struct tcp_seg {
 #define TF_SEG_OPTS_ZEROCOPY  (u8_t) TCP_WRITE_ZEROCOPY /* Use zerocopy send mode */
 
     u8_t tcp_flags; /* Cached TCP flags for outgoing segments */
+    u8_t bufs; /* To avoid pbuf_clen() */
```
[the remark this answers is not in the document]
alexbriskin: With the feedback from @AlexanderGrissik  I'll split the support for bufs into a separate PR.

src/core/sock/sockinfo_tcp.cpp
alexbriskin: Done

src/core/sock/sockinfo_tcp.cpp
alexbriskin: It is simpler to set the socket/connection state to fail/error.

src/core/lwip/tcp_in.c
alexbriskin: Will be implemented in a different PR.

src/core/sock/sockinfo_tcp.cpp
alexbriskin: done

src/core/sock/sockinfo_tcp.cpp
alexbriskin: Removing this expression in the slow_path might be better since we can allow for partial sends here regardless of the TCP window.

src/core/sock/sockinfo_tcp.cpp
alexbriskin: OK

src/core/sock/sockinfo_tcp.cpp
pasis: This comment is misplaced - it's actually related to the tcp_write_express() call above. We must guarantee that either full iov is accepted or not accepted at all. So, we need to do that around tcp_write_express(). Once all the iov is within the stack, tcp_output() may fail and progress will be triggered later by either further send(), RX or RTO.

I see 2 possible solutions for this:
1. push iov to tcp_write_express() instead of single pointers. In such a way tcp_write_express() will fail on memory error and no revert will be needed.
2. in corner case with memory error, revert partial data as a slow path.

I'd prefer (1) if it doesn't make impact on performance. Don't do this in context of current PR. Probably it will be done in context of aes_xts

src/core/sock/sockinfo_tcp.cpp
alexbriskin: It might be more correct here to use the is_connected_and_ready_to_send the same as in regular tcp_tx.
```
if (unlikely(!is_connected_and_ready_to_send())) {                                                                                                                                                                                                                                                                                  
    return -1;
}
```

src/core/sock/sockinfo_tcp.cpp
alexbriskin: ok

src/core/sock/sockinfo_tcp.cpp
alexbriskin: > tcp_output() doesn't have to guarantee to send anything to wire

src/core/sock/sockinfo_tcp.cpp
alexbriskin: In the current implementation, we first submit the entire iov to LwIP, and then we attempt tcp_output.
Moving the for loop (where we submit all the segments one-by-one) from sockinfo_tcp.cpp to tcp_out.c is not a big deal, as another call site already relies on a single segment press write API.

alexbriskin:
> seg->nbufs should be checked for correctness with each tcp_create_segment and currently not sure it will be handled well in all cases. For this PR we better remove this change and do it as sperate PR putting focus into this.

This won't be a part of this PR

src/core/sock/sockinfo_tcp.cpp
pasis: tcp_wirte-like functions prepare all the TCP segments and only in case of no errors do actual insert into the stream and update the PCB. So, moving the iov loop will resolve the issue with partial write, since tcp_write() already has the revert mechanism in case of errors. But again, this is in plans for other PR.

src/core/lwip/tcp_in.c
AlexanderGrissik: How do we handle here wrap arounds? even before the change?

src/core/sock/sockinfo_tcp.cpp
```diff
@@ -6046,3 +6055,178 @@ inline bool sockinfo_tcp::handle_bind_no_port(int &bind_ret, in_port_t in_port,
...
+    err_t err;
+    for (unsigned i = 0; i < iov_len; ++i) {
+        err = tcp_write_express(&m_pcb, iov[i].iov_base, iov[i].iov_len, &mdesc);
+        if (err != ERR_OK) {
+            /* The only error in tcp_write_express is a memory error
+             * In this version we don't implement any error recovery or avoidance
+             * mechanism and an error at this stage is irrecoverable.
+             * The considered alternatives are:
+             *  - Setting the socket an error state (this is the one we chose here)
+             *  - Rolling back any written buffers, i.e. recovering
```
AlexanderGrissik, line 6094: It seems that tcp_write_express already reverts back the segments on memory issue. So i guess we could just return error.

src/core/sock/sockinfo_ulp.cpp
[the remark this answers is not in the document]
AlexanderGrissik: I didn't notice comments about sndbuf overflow...

AlexanderGrissik (approved):

PR #71, opened 2023-12-07
issue: 3698256 Socket isolation feature

pasis:
## Description
Add socket isolation option to setsockopt() which allows XLIO to isolate specific sockets with some policy.

Currently, XLIO supports SO_XLIO_ISOLATE_SAFE logic which isolates sockets with a separate global thread-safe ring per interface. This allows to work with the sockets in multiple threads in a thread-unsafe configuration (e.g. delegated TCP timers mode).

In the future, SO_XLIO_ISOLATE interface can be extended with other logic.

##### What
Add socket isolation option to force thread-safety for specific sockets.

##### Why ?
Avoid races in thread-unsafe environment when some sockets need to be used in multiple threads.

## Change type
- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list

### DIFF
--- a/src/core/sock/sockinfo_tcp.cpp
+++ b/src/core/sock/sockinfo_tcp.cpp
@@ -4556,6 +4556,40 @@ int sockinfo_tcp::tcp_setsockopt(int __level, int __optname, __const void *__opt
             ret = -1;
             errno = EINVAL;
             break;
+        case SO_XLIO_ISOLATE:
+            // See option description in the extra API header.
+            pass_to_os_cond = false;
+            // We support SO_XLIO_ISOLATE only when no TX/RX rings are assigned.
+            if (__optlen == sizeof(int) && !m_p_connected_dst_entry && m_rx_ring_map.empty()) {
+                int tempval = *reinterpret_cast<const int *>(__optval);
+                bool ring_isolated =
+                    m_ring_alloc_log_rx.get_ring_alloc_logic() == RING_LOGIC_ISOLATE &&
+                    m_ring_alloc_log_tx.get_ring_alloc_logic() == RING_LOGIC_ISOLATE;
+
+                if (tempval == SO_XLIO_ISOLATE_DEFAULT && !ring_isolated) {
+                    // Do nothing.
+                    break;
+                }
+                if (tempval == SO_XLIO_ISOLATE_SAFE) {
+                    if (safe_mce_sys().tcp_ctl_thread ==
+                            option_tcp_ctl_thread::CTL_THREAD_DELEGATE_TCP_TIMERS &&
+                        !ring_isolated) {
+                        m_tcp_con_lock = multilock::create_new_lock(MULTILOCK_RECURSIVE, "tcp_con");
+                    }
+                    m_ring_alloc_log_rx = ring_alloc_logic_attr(RING_LOGIC_ISOLATE, true);
+                    m_ring_alloc_log_tx = ring_alloc_logic_attr(RING_LOGIC_ISOLATE, true);
+                    m_ring_alloc_logic_rx =
+                        ring_allocation_logic_rx(get_fd(), m_ring_alloc_log_rx, this);
+                    m_p_socket_stats->ring_alloc_logic_rx =
+                        m_ring_alloc_log_rx.get_ring_alloc_logic();
+                    m_p_socket_stats->ring_user_id_rx =
+                        m_ring_alloc_logic_rx.calc_res_key_by_logic();
+                    break;
+                }
+            }
+            ret = -1;
+            errno = EINVAL;
+            break;
         default:
             pass_to_os_always = true;
             supported = false;
--- a/tests/extra_api/socket_isolation.c
+++ b/tests/extra_api/socket_isolation.c
@@ -1,0 +1,277 @@
+/*
+ * Copyright © 2019-2023 NVIDIA CORPORATION & AFFILIATES. ALL RIGHTS RESERVED.
+ *
+ * This software is available to you under a choice of one of two
+ * licenses.  You may choose to be licensed under the terms of the GNU
+ * General Public License (GPL) Version 2, available from the file
+ * COPYING in the main directory of this source tree, or the
+ * BSD license below:
+ *
+ *     Redistribution and use in source and binary forms, with or
+ *     without modification, are permitted provided that the following
+ *     conditions are met:
+ *
+ *      - Redistributions of source code must retain the above
+ *        copyright notice, this list of conditions and the following
+ *        disclaimer.
+ *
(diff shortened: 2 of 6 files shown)


tests/extra_api/socket_isolation.c
```diff
@@ -0,0 +1,277 @@
...
+        server(argv[2]);
+    } else if (strcmp(argv[1], "-c") == 0) {
+        client(argv[2]);
+    } else {
+        usage(argv[0]);
+    }
+
+    printf("Success.\n");
+    return 0;
+}
```
alexbriskin, line 277: Why isn't this a part of our GTEST tests?

src/core/sock/sockinfo_tcp.cpp
```diff
@@ -4556,6 +4556,42 @@ int sockinfo_tcp::tcp_setsockopt(int __level, int __optname, __const void *__opt
...
+
+                    if (tempval == SO_XLIO_ISOLATE_DEFAULT && !ring_isolated) {
+                        /* Do nothing. */
+                        break;
+                    }
+                    if (tempval == SO_XLIO_ISOLATE_SAFE) {
+                        if (safe_mce_sys().tcp_ctl_thread ==
+                                option_tcp_ctl_thread::CTL_THREAD_DELEGATE_TCP_TIMERS &&
+                            !ring_isolated) {
+                            /* XXX TODO Is it correct way? */
```
alexbriskin, line 4576: What are the requirements for correct or incorrect?
How can we validate/test for "correctness"?

tests/extra_api/socket_isolation.c
pasis: manual test was easier to debug. This needs to be converted to a gtest for CI

src/core/sock/sockinfo_tcp.cpp
pasis: There are 2 requirements here:
1. We need to replace one lock object with another. From C++ perspective, is it optimal way?
2. We must (a) serialize lock replacement with the previous lock usage and (b) guarantee that the previous lock is not locked.

2(a) and 2(b) are guaranteed if user calls setsockopt() from the native thread (taking into account that this is branch for "TCP delegated timers").

src/core/sock/sockinfo_tcp.cpp
```diff
@@ -4556,6 +4556,42 @@ int sockinfo_tcp::tcp_setsockopt(int __level, int __optname, __const void *__opt
...
+                    bool ring_isolated =
+                        m_ring_alloc_log_rx.get_ring_alloc_logic() == RING_LOGIC_ISOLATE &&
+                        m_ring_alloc_log_tx.get_ring_alloc_logic() == RING_LOGIC_ISOLATE;
+
+                    if (tempval == SO_XLIO_ISOLATE_DEFAULT && !ring_isolated) {
+                        /* Do nothing. */
+                        break;
+                    }
+                    if (tempval == SO_XLIO_ISOLATE_SAFE) {
+                        if (safe_mce_sys().tcp_ctl_thread ==
```
alexbriskin, line 4573: Have you considered reducing the nesting by combining the two conditions?

tests/extra_api/socket_isolation.c
alexbriskin: Will you convert it as part of this PR?

src/core/sock/sockinfo_tcp.cpp
alexbriskin: It seems that this operation is inherently unsafe.
The locking and unlocking is a two-step operation:

1. unique_ptr dereference
2. locking the dereferenced lock pointer

The problematic scenario may happen when:

- Two threads may concurrently de-reference the pointer, and one may obtain the lock and replace the pointer.
- After the first thread releases/replaces the lock, the second thread may attempt to use a dangling lock pointer.

The **solution** might require a lock-replacing lock. To make the lock swapping a thread-safe operation, we might need the `mulilock::lock` operation to obtain the lock-replacing lock as well. The performance impact might be significant.

tests/extra_api/socket_isolation.c
pasis: lets review without gtest for now. I will convert it either in this PR or a separated depending on when the main logic is ready for merge

src/core/sock/sockinfo_tcp.cpp
```diff
@@ -4556,6 +4556,42 @@ int sockinfo_tcp::tcp_setsockopt(int __level, int __optname, __const void *__opt
...
+        case SO_XLIO_ISOLATE:
+            pass_to_os_cond = false;
+            if (__optlen == sizeof(int)) {
+                if (!m_p_connected_dst_entry && m_rx_ring_map.empty()) {
+                    int tempval = *reinterpret_cast<const int *>(__optval);
+                    bool ring_isolated =
+                        m_ring_alloc_log_rx.get_ring_alloc_logic() == RING_LOGIC_ISOLATE &&
+                        m_ring_alloc_log_tx.get_ring_alloc_logic() == RING_LOGIC_ISOLATE;
+
+                    if (tempval == SO_XLIO_ISOLATE_DEFAULT && !ring_isolated) {
```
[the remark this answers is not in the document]
pasis: suggested condition will break (and return success) in case of SO_XLIO_ISOLATE_DEFAULT **and** ring_isolated. But we must return EINVAL error in this case.

Since this is a control flow and proper application will unlikely call few times SO_XLIO_ISOLATE_SAFE in a row, lets keep logic as is and execute the branch with no useful effect. I don't see how to return success in this flow without complicating the code.

src/core/sock/sockinfo_tcp.cpp
```diff
@@ -4556,6 +4556,42 @@ int sockinfo_tcp::tcp_setsockopt(int __level, int __optname, __const void *__opt
             ret = -1;
             errno = EINVAL;
             break;
+        case SO_XLIO_ISOLATE:
+            pass_to_os_cond = false;
+            if (__optlen == sizeof(int)) {
+                if (!m_p_connected_dst_entry && m_rx_ring_map.empty()) {
```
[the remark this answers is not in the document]
pasis: these 2 ifs can be combined to reduce nesting. However I would prefer to avoid early break, because currently, all the failure paths are handled at the end when EINVAL is returned and all the success paths finish with a nested break. Handling EINVAL in the early break branch will overcomplicate code in my opinion.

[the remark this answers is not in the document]
pasis: combined 2 ifs and added a comment regarding logic of checking dst entry and ring_map

alexbriskin (approved):

PR #78, opened 2023-12-28
issue: 3613619 Avoid posting RX WQEs for Neigh ring

AlexanderGrissik:
## Description
Neigh ring was used only for TX and wasted RX buffers.

##### What
For Neigh ring we mark this ring as TX only and when QP is created RX WQEs are not posted to it's RQ.
This allows saving memory especially for multi process use cases.

##### Why ?
Avoid waste of memory.

## Change type
- [ ] Bugfix
- [X] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list


AlexanderGrissik:
I called this RING_ETH_TX. Let's discuss later if we want to change it to RING_ETH_NEIGH..

src/core/util/xlio_stats.h
```diff
@@ -351,9 +351,9 @@ typedef struct {
     cq_stats_t cq_stats;
 } cq_instance_block_t;
 
-typedef enum { RING_ETH = 0, RING_TAP } ring_type_t;
+typedef enum { RING_ETH = 0, RING_TAP, RING_ETH_TX } ring_type_t;
```
pasis, line 354: order ETH, ETH_TX, TAP looks more readable (if you change order here, change it for the strings array too)

src/core/xlio_extra.h
```diff
@@ -271,7 +271,7 @@ typedef enum {
     RING_LOGIC_PER_THREAD = 20, //!< RING_LOGIC_PER_THREAD
     RING_LOGIC_PER_CORE = 30, //!< RING_LOGIC_PER_CORE
     RING_LOGIC_PER_CORE_ATTACH_THREADS = 31, //!< RING_LOGIC_PER_CORE_ATTACH_THREADS
-    RING_LOGIC_PER_OBJECT = 32, //!< RING_LOGIC_PER_OBJECT
+    RING_LOGIC_NEIGH = 32, //!< RING_LOGIC_NEIGH
     RING_LOGIC_ISOLATE = 33, //!< RING_LOGIC_ISOLATE
     RING_LOGIC_LAST //!< RING_LOGIC_LAST
```
pasis, line 276: [suggestion]
xlio_extra.h:
```C
enum {
    ...
    RING_LOGIC_PER_CORE_ATTACH_THREADS = 31,
    RING_LOGIC_LAST = 100
};
```
ring_allocation_logic.h:
```C
/* Internal ring allocation logic which is hidden from user. */
enum {
    RING_LOGIC_NEIGH = RING_LOGIC_LAST + 1,
    RING_LOGIC_ISOLATE,
};
```
ring_allocation_logic.cpp:
```C
bool is_logic_support_migration()
{
...
    m_res_key.get_ring_alloc_logic() < **RING_LOGIC_LAST**
}
```

AlexanderGrissik: I am not sure about this change. It will create confusing conversions between ring_logic_t and ring_logic_public_t.
If we have to hide internal logics from the user of xlio_extra maybe it is better:
move ring_logic_t to ring_allocation_logic.h and make 
enum {
     ...
    RING_LOGIC_PER_CORE_ATTACH_THREADS = 31
} ring_logic_public_t

just for extra_api.h usage. The negative of this approach is that we have kind of duplicate definitions.
The best would be deriving one enum from another, but this is not possible in CPP.

src/core/util/xlio_stats.h
AlexanderGrissik: Done

AlexanderGrissik:
I rebased on top of vNext.

src/core/xlio_extra.h
pasis: this is extra API and will be replaced by storage api with the groups. Depending on further steps, we may even remove all the obsolete extra API. So lets keep it as is and improve once we do cleanup.

src/core/dev/hw_queue_rx.cpp
```diff
@@ -54,13 +54,14 @@
...
 hw_queue_rx::hw_queue_rx(ring_simple *ring, ib_ctx_handler *ib_ctx,
-                         ibv_comp_channel *rx_comp_event_channel, uint16_t vlan)
+                         ibv_comp_channel *rx_comp_event_channel, uint16_t vlan, bool dummy)
     : m_p_ring(ring)
     , m_p_ib_ctx_handler(ib_ctx)
     , m_n_sysvar_rx_num_wr_to_post_recv(safe_mce_sys().rx_num_wr_to_post_recv)
     , m_rx_num_wr(align32pow2(safe_mce_sys().rx_num_wr))
     , m_n_sysvar_rx_prefetch_bytes_before_poll(safe_mce_sys().rx_prefetch_bytes_before_poll)
     , m_vlan(vlan)
+    , m_dummy(dummy)
```
pasis, line 64: Would be better to make decision on dummy attribute in the constructor instead of storing it (lets say pass the information about dummy or number of buffers to `confiure_rq()`). If the RQ is really dummy, it won't be used and won't trigger any methods to post new buffers to the RQ.

If this cannot be done in context of constructor, then leave it in current state.

AlexanderGrissik:
The right approach is to split ring to ring_tx and ring_rx. Closing this PR as this does not resolve anything serious.

PR #81, opened 2024-01-01
Add support for static compilation with LTO and PGO

alexbriskin:
## Description
Add support for static compilation with LTO and PGO
**Enable LTO by default.**

## Change type
- [ ] Bugfix
- [X] Feature
- [ ] Code style update
- [x] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list

### DIFF
--- a/config/m4/compiler.m4
+++ b/config/m4/compiler.m4
@@ -71,107 +71,12 @@ AC_DEFUN([CHECK_COMPILER_ATTRIBUTE], [
 # Usage: CHECK_COMPILER_CXX([standard], [option], [definition])
 # Note:
 # - [definition] can be omitted if it is equal to attribute
-#
-AC_DEFUN([CHECK_COMPILER_CXX], [
-    case "$1" in
-        11)
-m4_define([_prj_cv_compiler_body_11], [[
-#ifndef __cplusplus
-#error This is not a C++ compiler
-#elif __cplusplus < 201103L
-#error This is not a C++11 compiler
-#else
-#include <iostream>
-int main(int argc, char** argv)
-{
-    (void)argc;
-    (void)argv;
-    /* decltype */
-    int a = 5;
-    decltype(a) b = a;
-    return (b - a);
-}
-#endif  // __cplusplus >= 201103L
-]])
-            ;;
-        14)
-m4_define([_prj_cv_compiler_body_14], [[
-#ifndef __cplusplus
-#error This is not a C++ compiler
-#elif __cplusplus < 201402L
-#error This is not a C++14 compiler
-#else
-#include <iostream>
-int main(int argc, char** argv)
-{
-    (void)argc;
-    (void)argv;
-    /* Binary integer literals */
-    constexpr auto i = 0b0000000000101010;
-    static_assert(i == 42, "wrong value");
-    return 0;
-}
-#endif  // __cplusplus >= 201402L
-]])
-            ;;
-        17)
-m4_define([_prj_cv_compiler_body_17], [[
-#ifndef __cplusplus
-#error This is not a C++ compiler
-#elif __cplusplus < 201703L
-#error This is not a C++17 compiler
-#else
-int main(int argc, char** argv)
-{
-    (void)argc;
-    (void)argv;
-    // Check constexpr lambda
-    auto identity = [](int n) constexpr { return n; };
-    static_assert(identity(123) == 123);
-    return 0;
-}
-#endif  // __cplusplus >= 201703L
-]])
-            ;;
-        *)
-            AC_MSG_ERROR([invalid first argument as [$1] to [$0]])
-            ;;
-    esac
-    case "$2" in
-        std)
-            prj_cv_option=-std=c++$1
-            ;;
-        gnu)
-            prj_cv_option=-std=gnu++$1
-            ;;
-        *)
-            AC_MSG_ERROR([invalid first argument as [$2] to [$0]])
-            ;;
-    esac
-
-    AC_CACHE_VAL(prj_cv_compiler_cxx_[$1], [
-        prj_cv_compiler_save_CXXFLAGS="$CXXFLAGS"
-        CXXFLAGS="$prj_cv_option $CXXFLAGS"
-
-        #
-        # Try to compile using the C++ compiler
-        #
-        AC_LANG_PUSH(C++)
-        AC_COMPILE_IFELSE([AC_LANG_SOURCE(_prj_cv_compiler_body_[$1])],
-                       [prj_cv_compiler_cxx_$1=yes],
-                       [prj_cv_compiler_cxx_$1=no])
-        AC_LANG_POP(C++)
-
-        CXXFLAGS="$prj_cv_compiler_save_CXXFLAGS"
-    ])
-    AC_MSG_CHECKING([for compiler c++ [$1]])
-    AC_MSG_RESULT([$prj_cv_compiler_cxx_$1])
-    AS_IF([test "x$prj_cv_compiler_cxx_[$1]" = "xyes"],
-        [CXXFLAGS="$prj_cv_option $CXXFLAGS"],
-        [AC_MSG_ERROR([A compiler with support for C++[$1] language features is required])]
-    )
-])
-
+saved_cxxflags="$CXXFLAGS"
+CXXFLAGS="-Werror -std=c++14"
+AC_MSG_CHECKING([whether CXX supports -std=c++14])
+AC_COMPILE_IFELSE([AC_LANG_PROGRAM([])], [AC_MSG_RESULT([yes])],
+	[AC_MSG_ERROR([C++14 is unsupported])])
+CXXFLAGS="-std=c++14 $saved_cxxflags"
 
 ##########################
 # Configure compiler capabilities
@@ -256,6 +161,4 @@ else
     AC_MSG_CHECKING([for symbols visibility])
     AC_MSG_RESULT([yes])
 fi
-
-CHECK_COMPILER_CXX([14], [std], [])
 ])
--- a/configure.ac
+++ b/configure.ac
@@ -105,9 +105,22 @@ show_section_title "Configure build tools"
 : ${CFLAGS=""}
 : ${CXXFLAGS=""}
 
+m4_include([config/m4/linking_optimization.m4])
+
 # Find compiler, libtools, etc
 #
 LT_INIT([disable-static])
+
+# LT_INIT exposes the ability to configure --enable-static
+if test "x$enable_static" = "xyes"; then
+	AC_SUBST([XLIO_STATIC_BUILD], ["-DXLIO_STATIC_BUILD"])
+	if test "x$enable_shared" = "xyes"; then
+		AC_MSG_ERROR([Please add --disable-shared or --enable-shared=no])
+	fi
+else
+	AC_SUBST([XLIO_STATIC_BUILD], [""])
+fi
+
 AC_PROG_CC
 AC_PROG_CXX
 
--- a/src/core/Makefile.am
+++ b/src/core/Makefile.am
@@ -12,7 +12,7 @@ LEX_OUTPUT_ROOT=lex.libxlio_yy
 # as built) because we don't want it to be created by old version of flex/yacc
 # on some machines that will generate gcc warmings.
 # in case you change the *.l or *.y in the future - than change the commenting in the following 3 lines
-#-----
+#
 #BUILT_SOURCES += config_scanner.c config_parser.h config_parser.c
 #libconfig_parser_la_SOURCES += util/config_scanner.l util/config_parser.y
 libconfig_parser_la_SOURCES += config_scanner.c config_parser.c
@@ -29,26 +29,32 @@ EXTRA_DIST = \
 
 sysconf_DATA = util/libxlio.conf
 otherincludedir = $(includedir)/mellanox
-otherinclude_HEADERS = xlio_extra.h
+otherinclude_HEADERS = xlio_extra.h \
+		       xlio.h
 
 install-exec-hook:
 	rm -f $(DESTDIR)$(libdir)/libxlio.la
-	rm -f $(DESTDIR)$(libdir)/libxlio.a
 	rm -f $(DESTDIR)$(bindir)/state_machine_test
 	rm -f $(DESTDIR)$(bindir)/vlogger_test
 
+
 uninstall-hook:
 	rm -f $(DESTDIR)$(libdir)/libxlio.so*
+	rm -f $(DESTDIR)$(libdir)/libxlio.a
 
 lib_LTLIBRARIES = libxlio.la
 
 AM_CPPFLAGS := \
 	-I$(top_srcdir)/src ${LIBNL_CFLAGS}
 
-libxlio_la_LDFLAGS := -no-undefined -version-number @PRJ_LIBRARY_MAJOR@:@PRJ_LIBRARY_MINOR@:@PRJ_LIBRARY_REVISION@ 
+libxlio_la_CFLAGS = $(XLIO_STATIC_BUILD) $(XLIO_LTO) $(XLIO_PROFILE)
+libxlio_la_CXXFLAGS = $(XLIO_STATIC_BUILD) $(XLIO_LTO) $(XLIO_PROFILE)
+
+libxlio_la_LDFLAGS := $(XLIO_LTO) $(XLIO_PROFILE) -no-undefined \
+	-version-number @PRJ_LIBRARY_MAJOR@:@PRJ_LIBRARY_MINOR@:@PRJ_LIBRARY_REVISION@
 
 libxlio_la_LIBADD = \
-	-lrt -ldl -lpthread $(LIBNL_LIBS) $(VERBS_LIBS) $(DPCP_LIBS) \
+	-lrt -ldl -lpthread $(LIBNL_LIBS) $(VERBS_LIBS) $(DPCP_LIBS) $(XLIO_GCOV) \
 	$(top_builddir)/src/utils/libutils.la \
 	$(top_builddir)/src/vlogger/libvlogger.la \
 	$(top_builddir)/src/state_machine/libstate_machine.la \
--- a/src/core/sock/sock-app.cpp
+++ b/src/core/sock/sock-app.cpp
@@ -35,6 +35,7 @@
 #endif
 
 #include <sock/socket_fd_api.h>
+#include <sock/sock-redirect.h>
 #include <sock/sockinfo_tcp.h>
 #include <sock/sockinfo_udp.h>
 #include <sock/fd_collection.h>
@@ -253,7 +254,7 @@ static int init_worker(int worker_id, int listen_fd)
         if (child_sock_fd_api) {
             child_sock_fd_api->copy_sockopt_fork(parent_sock_fd_api);
 
-            ret = bind(listen_fd, sa.get_p_sa(), sa_len);
+            ret = bind_internal(child_sock_fd_api, sa.get_p_sa(), sa_len);
             if (ret < 0) {
                 app_logerr("bind() error");
             }
--- a/src/core/sock/sock-extra.cpp
+++ b/src/core/sock/sock-extra.cpp
@@ -83,12 +83,7 @@ extern "C" int xlio_recvfrom_zcopy(int __fd, void *__buf, size_t __nbytes, int *
         *__flags |= MSG_XLIO_ZCOPY;
         return p_socket_object->rx(RX_RECVFROM, piov, 1, __flags, __from, __fromlen);
     }
-    BULLSEYE_EXCLUDE_BLOCK_START
-    if (!orig_os_api.recvfrom) {
-        get_orig_funcs();
-    }
-    BULLSEYE_EXCLUDE_BLOCK_END
-    return orig_os_api.recvfrom(__fd, __buf, __nbytes, *__flags, __from, __fromlen);
+    return SYSCALL(recvfrom, __fd, __buf, __nbytes, *__flags, __from, __fromlen);
 }
 
 extern "C" int xlio_recvfrom_zcopy_free_packets(int __fd, struct xlio_recvfrom_zcopy_packet_t *pkts,
@@ -315,7 +310,7 @@ static inline struct cmsghdr *__cmsg_nxthdr(void *__ctl, size_t __size, struct c
     return __ptr;
 }
 
-extern "C" int xlio_ioctl(void *cmsg_hdr, size_t cmsg_len)
+extern "C" int xlio_extra_ioctl(void *cmsg_hdr, size_t cmsg_len)
 {
     struct cmsghdr *cmsg = (struct cmsghdr *)cmsg_hdr;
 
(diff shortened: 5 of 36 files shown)


configure.ac
```diff
@@ -107,7 +108,17 @@ show_section_title "Configure build tools"
 
 # Find compiler, libtools, etc
 #
-LT_INIT([disable-static])
+AC_ARG_ENABLE([xlio-static-lib],
```
pasis, line 111: I'd prefer to use --enable-static which comes out of box with LT_INIT and does the same logic. In this case LT_INIT([disable-static]) should be untouched and user may request --enable-static manually. We can pass this information further to c++ code via define to handle syscall symbols.

src/core/Makefile.am
```diff
@@ -35,12 +35,16 @@ otherinclude_HEADERS = xlio_extra.h
 
 install-exec-hook:
 	rm -f $(DESTDIR)$(libdir)/libxlio.la
-	rm -f $(DESTDIR)$(libdir)/libxlio.a
 	rm -f $(DESTDIR)$(bindir)/state_machine_test
 	rm -f $(DESTDIR)$(bindir)/vlogger_test
+if DISABLE_XLIO_STATIC_LIB
+	rm -f $(DESTDIR)$(libdir)/libxlio.a
+endif
```
pasis, line 42: Once default `--enable-static` is used, this block should be removed, because user will control whether static library is built or not.

configure.ac
alexbriskin: Please suggest a fix.

src/core/Makefile.am
alexbriskin: None of the experiments I did with --enable-static would work. Please suggest a fix.

configure.ac
alexbriskin: Done

src/core/Makefile.am
alexbriskin: Done

src/core/xlio_extra.h
```diff
@@ -641,4 +645,163 @@ static inline struct xlio_api_t *xlio_get_api()
...
+struct mmsghdr {
+    struct msghdr msg_hdr; // Message header
+    unsigned int msg_len; // Number of received bytes for header
+};
+#endif
+#endif
+
+#ifndef MSG_WAITFORONE
+#define MSG_WAITFORONE 0x10000 // recvmmsg(): block until 1+ packets avail
+#endif
```
pasis, line 709: Please remove this definition.

src/core/xlio_extra.h
```diff
@@ -641,4 +645,163 @@ static inline struct xlio_api_t *xlio_get_api()
...
+
+#ifdef RECVMMSG_WITH_CONST_TIMESPEC
+int xlio_recvmmsg(int __fd, struct mmsghdr *__mmsghdr, unsigned int __vlen, int __flags,
+                  const struct timespec *__timeout);
+#else
+int xlio_recvmmsg(int __fd, struct mmsghdr *__mmsghdr, unsigned int __vlen, int __flags,
+                  struct timespec *__timeout);
+#endif
```
pasis, line 717: I'd prefer to keep a single prototype - user uses xlio extra API explicitly what doesn't have to follow specific libc implementation byte by byte. const version looks like version which will work in more cases.

Also, having such XLIO specific define in public header is error prone. AFAIK we don't expose this define to pkg-config file and there is no guarantee that user will use pkg-config.

src/core/xlio_extra.h
```diff
@@ -641,4 +645,163 @@ static inline struct xlio_api_t *xlio_get_api()
...
+
+int xlio_creat(const char *__pathname, mode_t __mode);
+
+int xlio_dup(int __fd);
+
+int xlio_dup2(int __fd, int __fd2);
+
+#ifdef _CHANGE_CLONE_PROTO_IN_SLES_10_
+int xlio_clone(int (*__fn)(void *), void *__child_stack, int __flags, void *__arg);
+#endif
```
pasis, line 789: Remove this function. I don't see any usage for it. Also, I'd prefer to remove its overloading from sock-redirect.cpp. Looks like the only purpose is to profile timing and print debug message when _CHANGE_CLONE_PROTO_IN_SLES_10_ is set manually and this define cannot be configured by user.

src/core/xlio_extra.h
```diff
@@ -641,4 +645,163 @@ static inline struct xlio_api_t *xlio_get_api()
     return api_ptr;
 }
 
+extern "C" {
+int xlio_socket(int __domain, int __type, int __protocol);
+
+int xlio_close(int __fd);
+
+void xlio___res_iclose(res_state statp, bool free_addr);
```
pasis, line 653: Remove this prototype. This is internal libresolv logic. Since user uses xlio_socket API explicitly, all dependent static/shared libraries will use systems sockets without offloading. So, libresolv will use kernel for socket interaction and libc __res_iclose() will close all internal sockets with kernel.

We have to overload this symbol in case of LD_PRELOAD, because XLIO may offload libresolv sockets in this scenario and kernel close() will break functionality.

src/core/xlio_extra.h
```diff
@@ -641,4 +645,163 @@ static inline struct xlio_api_t *xlio_get_api()
...
+int xlio_sendmmsg(int __fd, struct mmsghdr *__mmsghdr, unsigned int __vlen, int __flags);
+
+ssize_t xlio_sendto(int __fd, __const void *__buf, size_t __nbytes, int __flags,
+                    const struct sockaddr *__to, socklen_t __tolen);
+
+ssize_t xlio_sendfile(int out_fd, int in_fd, off_t *offset, size_t count);
+
+ssize_t xlio_sendfile64(int out_fd, int in_fd, __off64_t *offset, size_t count);
+
+const char *dbg_sprintf_fdset(char *buf, int buflen, int __nfds, fd_set *__fds);
```
pasis, line 744: remove this prototype

src/core/xlio_extra.h
```diff
@@ -641,4 +645,163 @@ static inline struct xlio_api_t *xlio_get_api()
...
+
+int xlio_sigaction(int signum, const struct sigaction *act, struct sigaction *oldact);
+
+sighandler_t xlio_signal(int signum, sighandler_t handler);
+
+#if defined(DEFINED_NGINX)
+int xlio_setuid(uid_t uid);
+
+pid_t xlio_waitpid(pid_t pid, int *wstatus, int options);
+#endif // DEFINED_NGINX
```
pasis, line 805: Remove this block. NGINX doesn't use XLIO explicitly.

src/core/xlio_extra.h
```diff
@@ -37,6 +37,8 @@
 #include <stdint.h>
 #include <netinet/in.h>
 #include <sys/socket.h>
+#include <resolv.h>
```
pasis, line 40: remove resolv.h -- look like it's not needed.

src/core/sock/sock-redirect.h
```diff
@@ -85,6 +89,14 @@
 #include <mutex>
 #include <set>
 
+#ifdef XLIO_STATIC_BUILD
+#define DEF_SYM(_name)     xlio_##_name
+#define ORIG_SYMBOL(_name) _name
+#else
+#define DEF_SYM(_name)     _name
+#define ORIG_SYMBOL(_name) orig_os_api._name
+#endif /* XLIO_STATIC_BUILD */
```
pasis, line 98: [suggestion] probably name with "SYSCALL" will be more readable. For example, glibc has something like this: `INTERNAL_SYSCALL_CALL`, `INLINE_SYSCALL`. With "INTERNAL_SYSCALL" or "XLIO_SYSCALL" we can call XLIO implementation and "EXTERNAL_SYSCALL" or "KERNEL_SYSCALL" will replace ORIG_SYMBOL.

configure.ac
```diff
@@ -107,7 +108,16 @@ show_section_title "Configure build tools"
 
 # Find compiler, libtools, etc
 #
+
 LT_INIT([disable-static])
+
+# LT_INIT exposes the ability to configure --enable-static
```
pasis, line 114: Please add an error message if both shared and static libraries are requested. Current version won't build both binaries properly.

src/core/sock/sock-redirect.h
pasis: Also, need to choose between current format and the following one:
```C
#define DEF_SYM(_name, ...) xlio_##_name(__VA_ARGS__)
```
So, calling a syscall would look like DEF_SYM(close, fd)

src/core/xlio_extra.h
alexbriskin: It is needed for the __recv* functions

configure.ac
alexbriskin: Please elaborate, why wouldn't the current version build both binaries?

src/core/xlio_extra.h
pasis: you mean __recv_chk() and __recvfrom_chk()? resolv.h is not needed for them. It's purpose is to define `res_state` for __res_iclose() which should be removed from xlio_extra.h. I don't see other usage of resolv.h here.

configure.ac
pasis: because shared library will be build with XLIO_STATIC_BUILD define in this case and socket API will be replaced with xlio_*.

src/core/sock/sock-redirect.h
pasis: regarding naming: I like XLIO_CALL() more than XLIO(). In this case kernel syscall can follow this convention too - KERNEL_CALL(). Lets bring this discussion to wider forum.

Regarding VA_ARGS: these are just alternative options. I don't have strong opinion here. As for VA_ARGS option, this looks like commonly used approach and provides clear construction of function call, while `DEF_SYM(func)(args)` looks more complex.

src/core/xlio_extra.h
```diff
@@ -641,4 +645,163 @@ static inline struct xlio_api_t *xlio_get_api()
...
+
+int xlio_getsockname(int __fd, struct sockaddr *__name, socklen_t *__namelen);
+
+int xlio_getpeername(int __fd, struct sockaddr *__name, socklen_t *__namelen);
+
+ssize_t xlio_read(int __fd, void *__buf, size_t __nbytes);
+
+#if defined HAVE___READ_CHK
+ssize_t xlio___read_chk(int __fd, void *__buf, size_t __nbytes, size_t __buflen);
+#endif
```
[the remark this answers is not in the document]
alexbriskin: I can't do that safely as long as there is a definition that might statically call the original *_chk function

src/core/xlio_extra.h
alexbriskin: OK

src/core/xlio_extra.h
alexbriskin: OK

src/core/xlio_extra.h
alexbriskin: OK

src/core/xlio_extra.h
```diff
@@ -641,4 +645,146 @@ static inline struct xlio_api_t *xlio_get_api()
...
+
+pid_t xlio_fork(void);
+
+pid_t xlio_vfork(void);
+
+int xlio_daemon(int __nochdir, int __noclose);
+
+int xlio_sigaction(int signum, const struct sigaction *act, struct sigaction *oldact);
+
+sighandler_t xlio_signal(int signum, sighandler_t handler);
```
pasis, line 782: [Note for future] We should remove these set of prototypes from xlio_extra.h and provide API which solves the same problems: xlio_extra_init() / xlio_extra_terminate()

src/core/xlio_extra.h
alexbriskin: It would be shocking for SPDK or some other future client to statically link against XLIO and find out that we override the __res_iclose, and there is no way to opt out.

src/core/xlio_extra.h
alexbriskin: It would be shocking for SPDK or some other future client to statically link against XLIO and find out that we override the __res_iclose, and there is no way to opt out

config/m4/compiler.m4
```diff
@@ -71,107 +71,11 @@ AC_DEFUN([CHECK_COMPILER_ATTRIBUTE], [
...
-    AS_IF([test "x$prj_cv_compiler_cxx_[$1]" = "xyes"],
-        [CXXFLAGS="$prj_cv_option $CXXFLAGS"],
-        [AC_MSG_ERROR([A compiler with support for C++[$1] language features is required])]
-    )
-])
-
+saved_cxxflags="$CXXFLAGS"
+CXXFLAGS="-Werror -std=c++14"
+AC_MSG_CHECKING([whether CXX supports -std=c++14])
+AC_COMPILE_IFELSE([AC_LANG_PROGRAM([])], [AC_MSG_RESULT([yes])], [AC_MSG_RESULT([no])])
```
pasis, line 77: AFAIU, this will simply print whether -std=c++14 is supported or not and continue. But don't we expect to set the standard to CXXFLAGS and fail configure if the requested standard is not supported?

alexbriskin: I replaced AC_MSG_RESULT with AC_MSG_ERROR; the check should fail, and the configuration should stop.

src/core/xlio_extra.h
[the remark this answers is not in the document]
benlwalker: There won't be any static callers of the original *_chk functions - only glibc knows about them. They are not part of POSIX sockets.

src/core/xlio_extra.h
benlwalker: The two variations here are to support Linux (no const) and BSD (const). It shouldn't be selectable by the user - it should be determined by the build system. I think this has to remain here if XLIO supports BSD.

src/core/xlio_extra.h
benlwalker: Yes, long term what we really want is a brand new "static library" API with it's own header file that's designed to be efficient, and then separately the POSIX hooks only if you compile as an LD_PRELOADable library.

src/core/sock/sock-redirect.h
alexbriskin: OK

src/core/xlio_extra.h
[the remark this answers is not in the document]
alexbriskin: OK

src/core/xlio_extra.h
alexbriskin: OK

src/core/xlio_extra.h
alexbriskin: OK

src/core/xlio_extra.h
alexbriskin: OK

src/core/xlio_extra.h
alexbriskin: Done

src/core/sock/sock-app.cpp
```diff
@@ -253,7 +254,7 @@ static int init_worker(int worker_id, int listen_fd)
         if (child_sock_fd_api) {
             child_sock_fd_api->copy_sockopt_fork(parent_sock_fd_api);
 
-            ret = bind(listen_fd, sa.get_p_sa(), sa_len);
+            ret = bind_internal(child_sock_fd_api, sa.get_p_sa(), sa_len);
```
AlexanderGrissik, line 257: Shouldn't it be listen_fd?

src/core/sock/sock-extra.cpp
```diff
@@ -83,12 +83,7 @@ extern "C" int xlio_recvfrom_zcopy(int __fd, void *__buf, size_t __nbytes, int *
         *__flags |= MSG_XLIO_ZCOPY;
         return p_socket_object->rx(RX_RECVFROM, piov, 1, __flags, __from, __fromlen);
     }
-    BULLSEYE_EXCLUDE_BLOCK_START
-    if (!orig_os_api.recvfrom) {
```
AlexanderGrissik, line 87: Should we ignore this check?

configure.ac
alexbriskin: The shared library with enabled static is different from the one without it.
Do we need an additional flag to turn on/off the XLIO_STATIC_BUILD?

src/core/sock/sock-app.cpp
alexbriskin: No. Because of lines 253-254, the listen_fd and child_socket_fd_api reference the same object. The first reference is with a file descriptor, and the other is with a pointer to the object memory address.

src/core/sock/sock-extra.cpp
alexbriskin: Please referer https://github.com/alexbriskin/libxlio_public/blob/ac507222d23e75a0d0f09090eec186187a357ea3/src/core/sock/sock-redirect.h#L105

The answer is that the SYSCALL macro in the dynamic library case:

1. Checks the syscall initialized in the orig_os_api object
2. If not, initializes the object
3. Invoke the function pointer

In the static case, the SYSCALL macro calls the original symbols without the xlio_ prefix.

src/core/sock/sock-redirect.h
```diff
@@ -85,6 +89,26 @@
...
+#define SYSCALL_ERRNO_UNSUPPORTED(_func, ...) SYSCALL(_func, __VA_ARGS__)
+#define VALID_SYSCALL(_func)                  (true)
+#else
+#define XLIO_SYMBOL(_func) _func
+#if defined(__GNUC__) && !defined(__clang__)
+#define VALID_SYSCALL(_func) (__builtin_addressof(orig_os_api._func) != nullptr)
+#else
+#define VALID_SYSCALL(_func) ((orig_os_api._func) != nullptr)
+#endif
+#define SYSCALL(_func, ...)                                                                        \
```
AlexanderGrissik, line 105: In case VALID_SYSCALL(_func) is true why should we call get_orig_funcs()? shouldn't be the opposite?

src/core/util/agent.cpp
```diff
@@ -62,10 +62,7 @@
 /* Force system call */
 #define sys_call(_result, _func, ...)                                                              \
     do {                                                                                           \
-        if (orig_os_api._func)                                                                     \
-            _result = orig_os_api._func(__VA_ARGS__);                                              \
-        else                                                                                       \
-            _result = ::_func(__VA_ARGS__);                                                        \
+        _result = SYSCALL(_func, __VA_ARGS__);                                                     \
```
AlexanderGrissik, line 65: SYSCALL will not call the ::func version if orig funcs not found...

src/core/util/utils.cpp
```diff
@@ -674,20 +674,20 @@ class socket_context_manager {
...
-        m_fd = orig_os_api.socket(AF_NETLINK, SOCK_RAW, NETLINK_ROUTE);
+        m_fd = SYSCALL(socket, AF_NETLINK, SOCK_RAW, NETLINK_ROUTE);
         if (m_fd < 0) {
             throw std::runtime_error("Open netlink socket failed");
         }
 
-        if (orig_os_api.setsockopt(m_fd, SOL_SOCKET, SO_RCVTIMEO, (const char *)&tv, sizeof tv)) {
-            close(m_fd);
+        if (SYSCALL(setsockopt, m_fd, SOL_SOCKET, SO_RCVTIMEO, (const char *)&tv, sizeof tv)) {
+            SYSCALL(close, m_fd);
```
AlexanderGrissik, line 683: Shouldn't be XLIO_CALL?

src/core/xlio_extra.h
alexbriskin: I will leave the const variant since this is a new symbol.

configure.ac
alexbriskin: OK

src/core/xlio_extra.h
alexbriskin: Removed and added xlio_init/xlio_exit with the appropriate comments

src/core/sock/sock-redirect.h
alexbriskin: Yes, change is done.

src/core/util/agent.cpp
alexbriskin: That's an error.
I'll revert the code.

src/core/util/utils.cpp
alexbriskin: since we opened the socket with `orig_os_api.socket` / `SYSCALL` the expected behavior is to use a similar close.

src/core/util/utils.cpp
```diff
@@ -674,20 +674,20 @@ class socket_context_manager {
...
             throw std::runtime_error("Setsockopt non-blocking failed");
         }
     }
 
     socket_context_manager(int fd) noexcept
         : m_fd(fd) {};
-    ~socket_context_manager() { close(m_fd); };
+    ~socket_context_manager() { SYSCALL(close, m_fd); };
```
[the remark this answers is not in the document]
alexbriskin: Same as before, since we opened the socket with `orig_os_api.socket` / `SYSCAL`L the expected behavior is to use a similar close.

alexbriskin:
> Please add checking for LTO ... checking for PGO (profile/use) ... prints to the configure.

Done

AlexanderGrissik:
We need to disable generate-profile/use-profile for xlio_stats/daemon/other because if we do not run these tools to create profile files the use use stage compilation throws warnings.
`  CC       xliod-daemon.o
  CC       xliod-loop.o
  CC       xliod-store.o
  CC       xliod-flow.o
  CC       xliod-message.o
  CC       xliod-notify.o
../../../tools/daemon/store.c: In function ‘free_store_pid’:
../../../tools/daemon/store.c:72:1: warning: ‘/labhome/agrissik/mtrswgwork/xlio/snap_express/libxlio/out_bf3_0124/profile/#labhome#agrissik#mtrswgwork#xlio#snap_express#libxlio#out_bf3_0124#tools#daemon#xliod-store.gcda’ profile count data file not found [-Wmissing-profile]
   72 | }
      | ^
../../../tools/daemon/loop.c: In function ‘proc_loop’:
../../../tools/daemon/loop.c:143:1: warning: ‘/labhome/agrissik/mtrswgwork/xlio/snap_express/libxlio/out_bf3_0124/profile/#labhome#agrissik#mtrswgwork#xlio#snap_express#libxlio#out_bf3_0124#tools#daemon#xliod-loop.gcda’ profile count data file not found [-Wmissing-profile]
  143 | }
      | ^
../../../tools/daemon/daemon.c: In function ‘sys_exec’:
../../../tools/daemon/daemon.c:475:1: warning: ‘/labhome/agrissik/mtrswgwork/xlio/snap_express/libxlio/out_bf3_0124/profile/#labhome#agrissik#mtrswgwork#xlio#snap_express#libxlio#out_bf3_0124#tools#daemon#xliod-daemon.gcda’ profile count data file not found [-Wmissing-profile]
  475 | }
      | ^
../../../tools/daemon/message.c: In function ‘proc_msg_flow’:
../../../tools/daemon/message.c:535:1: warning: ‘/labhome/agrissik/mtrswgwork/xlio/snap_express/libxlio/out_bf3_0124/profile/#labhome#agrissik#mtrswgwork#xlio#snap_express#libxlio#out_bf3_0124#tools#daemon#xliod-message.gcda’ profile count data file not found [-Wmissing-profile]
  535 | }
      | ^
../../../tools/daemon/flow.c: In function ‘get_node’:
../../../tools/daemon/flow.c:771:1: warning: ‘/labhome/agrissik/mtrswgwork/xlio/snap_express/libxlio/out_bf3_0124/profile/#labhome#agrissik#mtrswgwork#xlio#snap_express#libxlio#out_bf3_0124#tools#daemon#xliod-flow.gcda’ profile count data file not found [-Wmissing-profile]
  771 | }
      | ^
../../../tools/daemon/notify.c: In function ‘proc_inotify’:
../../../tools/daemon/notify.c:1019:1: warning: ‘/labhome/agrissik/mtrswgwork/xlio/snap_express/libxlio/out_bf3_0124/profile/#labhome#agrissik#mtrswgwork#xlio#snap_express#libxlio#out_bf3_0124#tools#daemon#xliod-notify.gcda’ profile count data file not found [-Wmissing-profile]
 1019 | }
      | ^
`

AlexanderGrissik (approved):

PR #89, opened 2024-01-24
[CI] Coverity snapshots integration

vialogi:
We'd like to make CI able to post Coverity reports to https://coverity.mellanox.com/
[HPCINFRA-1430](https://jirasw.nvidia.com/browse/HPCINFRA-1430)


contrib/jenkins_tests/cov.sh
dpressle, line 1: Do you use shell check? it seems like you are not. for example, variables should be quoted

dpressle, line 1: We should use bash and not posix (as a team guide line that we want to follow)
if [[ $do_coverity_snapshot = true ]];

contrib/jenkins_tests/cov.sh
```diff
@@ -46,6 +46,15 @@ eval "cov-analyze --config ${cov_dir}/coverity_config.xml \
...
 
+if [ "${do_coverity_snapshot}" == "true" ]; then
+    cov-commit-defects --ssl --on-new-cert trust \
+        --url https://coverity.mellanox.com:8443 \
+        --user ${XLIO_COV_USER} --password ${XLIO_COV_PASSWORD} \
+        --dir "${cov_build}" \
+        --stream libxlio-main \
+        --strip-path ${WORKSPACE}
+fi
+
```
dpressle, line 57: what about failing the job if it didnt work ? like in line 47

dpressle (changes requested):
I also would like to see link to a test run in the comments

vialogi:
Sample test run:
http://hpc-master.lab.mtl.com:8080/job/LibXLIO-opensource/457/

contrib/jenkins_tests/cov.sh
```diff
@@ -46,6 +46,16 @@ eval "cov-analyze --config ${cov_dir}/coverity_config.xml \
...
 
+if [[ "${do_coverity_snapshot}" == true ]]; then
+    cov-commit-defects --ssl --on-new-cert trust \
+        --url https://coverity.mellanox.com:8443 \
+        --user "${XLIO_COV_USER}" --password "${XLIO_COV_PASSWORD}" \
+        --dir "${cov_build}" \
+        --stream libxlio-main \
+        --strip-path "${WORKSPACE}"
+fi
+rc=$(($rc+$?))
```
dpressle, line 57: It can be meaningless here (when we dont run cov commit for example), should be in the if/fi block

dpressle (changes requested):

contrib/jenkins_tests/cov.sh
dpressle: i dont see the fix ?

dpressle (changes requested):

contrib/jenkins_tests/cov.sh
```diff
@@ -46,6 +46,16 @@ eval "cov-analyze --config ${cov_dir}/coverity_config.xml \
...
 rc=$(($rc+$?))
 
+if [[ "${do_coverity_snapshot}" == true ]]; then
+    cov-commit-defects --ssl --on-new-cert trust \
+        --url https://coverity.mellanox.com:8443 \
+        --user "${XLIO_COV_USER}" --password "${XLIO_COV_PASSWORD}" \
+        --dir "${cov_build}" \
+        --stream libxlio-main \
+        --strip-path "${WORKSPACE}"
+rc=$(($rc+$?))
```
dpressle, line 56: Fix Indentation

dpressle (changes requested):

PR #90, opened 2024-01-28
issue: 3724170 Add missing ifdef __cplusplus

alexbriskin:
## Description
Add missing ifdef cplusplus
## Change type
- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list

### DIFF
--- a/contrib/jenkins_tests/build.sh
+++ b/contrib/jenkins_tests/build.sh
@@ -21,7 +21,7 @@ build_list['static-on']="--enable-static --disable-shared"
 build_list['default']=""
 
 build_tap=${WORKSPACE}/${prefix}/build.tap
-echo "1..$(echo $build_list | tr " " "\n" | wc -l)" > $build_tap
+echo "1..${#build_list[@]}" > $build_tap
 
 test_id=0
 
@@ -32,7 +32,7 @@ for build_name in "${!build_list[@]}"; do
     test_exec='${WORKSPACE}/configure --prefix=${build_dir}/${test_id}/install $build_option $jenkins_test_custom_configure && make $make_opt install'
     do_check_result "$test_exec" "$test_id" "$build_name" "$build_tap" "${build_dir}/build-${test_id}"
     cd ${build_dir}
-    ((test_id++))
+    test_id=$((test_id+1))
 done
 
 
--- a/contrib/jenkins_tests/build.sh
+++ b/contrib/jenkins_tests/build.sh
@@ -13,25 +13,26 @@ cd ${build_dir}
 # Set symbolic links to default build and install
 ln -s "${build_dir}/0/install" "${install_dir}"
 
-build_list="\
-debug:--enable-opt-log=no --enable-debug \
-nginx-off:--enable-nginx=no \
-envoy-on:--enable-nginx=yes \
-default: "
-
+declare -A build_list
+build_list['debug']="--enable-opt-log=no --enable-debug"
+build_list['nginx-off']="--enable-nginx=no"
+build_list['envoy-on']="--enable-nginx=yes"
+build_list['static-on']="--enable-static --disable-shared"
+build_list['default']=""
 
 build_tap=${WORKSPACE}/${prefix}/build.tap
 echo "1..$(echo $build_list | tr " " "\n" | wc -l)" > $build_tap
 
 test_id=0
-for build in $build_list; do
-    IFS=':' read build_name build_option <<< "$build"
+
+for build_name in "${!build_list[@]}"; do
+    build_option="${build_list[$build_name]}"
     mkdir -p ${build_dir}/${test_id}
     cd ${build_dir}/${test_id}
     test_exec='${WORKSPACE}/configure --prefix=${build_dir}/${test_id}/install $build_option $jenkins_test_custom_configure && make $make_opt install'
     do_check_result "$test_exec" "$test_id" "$build_name" "$build_tap" "${build_dir}/build-${test_id}"
     cd ${build_dir}
-    test_id=$((test_id+1))
+    ((test_id++))
 done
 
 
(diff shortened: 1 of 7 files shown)


src/core/xlio.h
```diff
@@ -36,6 +36,7 @@
 #include <signal.h>
 #include <sys/poll.h>
 
+#ifdef __cplusplus
```
benlwalker, line 39: This makes the whole API unavailable to C. What you want instead is this:

#ifdef __cplusplus
extern "C" {
#endif

And similarly below for the closing bracket.

benlwalker (changes requested):

src/core/sock/sock-redirect.h
```diff
@@ -207,8 +207,12 @@ struct os_api {
 #if defined(DEFINED_NGINX)
     int (*setuid)(uid_t uid);
     pid_t (*waitpid)(pid_t pid, int *wstatus, int options);
-#endif // DEFINED_NGINX
 };
+#endif /* XLIO_STATIC_BUILD */
+extern os_api orig_os_api;
+
+extern void get_orig_funcs();
+#endif // DEFINED_NGINX
```
pasis, line 215: Something wrong with the #endif placement and their comments

alexbriskin:
bot:retest

src/core/xlio.h
alexbriskin: OK

src/core/sock/sock-redirect.h
alexbriskin: OK

alexbriskin:
bot:retest

benlwalker (approved):

src/core/sock/sock-redirect.cpp
```diff
@@ -118,15 +119,6 @@ template <typename T> void assign_dlsym(T &ptr, const char *name)
...
-#define VERIFY_PASSTROUGH_CHANGED(__ret, __func_and_params__)                                      \
-    do {                                                                                           \
-        bool passthrough = p_socket_object->isPassthrough();                                       \
-        __ret = __func_and_params__;                                                               \
-        if (!passthrough && p_socket_object->isPassthrough()) {                                    \
-            handle_close(__fd, false, true);                                                       \
-        }                                                                                          \
-    } while (0);
-
 void get_orig_funcs()
```
benlwalker, line 122: Now that this is behind the XLIO_STATIC_BUILD flag, then xlio_init (which calls this) no longer compiles.

configure.ac
```diff
@@ -322,7 +322,7 @@ show_section_title "Check for functions, types and structures"
 
 # Check fortified glibc functions
 #
```
benlwalker, line 324: This still doesn't work correctly for me. Even in a static build, it decides that these functions exist here. But when I compile the static library I get errors like this:

sock/sock-redirect.cpp: In function ‘ssize_t xlio___read_chk(int, void*, size_t, size_t)’:
sock/sock-redirect.cpp:1365:20: error: ‘::__read_chk’ has not been declared
 1365 |     return SYSCALL(__read_chk, __fd, __buf, __nbytes, __buflen);

benlwalker (changes requested):

configure.ac
alexbriskin: Will enable the functions for non-static build

src/core/sock/sock-redirect.cpp
alexbriskin: Fixed

contrib/jenkins_tests/build.sh
```diff
@@ -13,19 +13,20 @@ cd ${build_dir}
...
-nginx-off:--enable-nginx=no \
-envoy-on:--enable-nginx=yes \
-default: "
-
+declare -A build_list
+build_list['debug']="--enable-opt-log=no --enable-debug"
+build_list['nginx-off']="--enable-nginx=no"
+build_list['envoy-on']="--enable-nginx=yes"
+build_list['static-on']="--enable-static --disable-shared"
+build_list['default']=""
```
pasis, line 21: I'd also prefer to see maximum build (nginx, envoy) and --enable-opt-log=high. "high" value is not covered by these builds and is a source of warnings that can be easily missed until the release build.

Not necessary in current PR.

src/core/sock/sock-redirect.cpp
```diff
@@ -1103,12 +1105,14 @@ EXPORT_SYMBOL int XLIO_SYMBOL(getsockopt)(int __fd, int __level, int __optname,
 
     srdr_logdbg_entry("fd=%d, level=%d, optname=%d", __fd, __level, __optname);
 
+#ifndef XLIO_STATIC_BUILD
```
pasis, line 1108: What happens with xlio_get_api() function which can be used from <xlio_extra.h> even with static linkage? I expect it will fail with static lib, because getsockopt() will go to kernel?

I guess not much we can do with current xlio_get_api() implementation...

pasis (approved):

contrib/jenkins_tests/build.sh
alexbriskin: This change is designed to fix the bug where build_list with multiple options was broken.
In addition, I added the test for static build to CI.
Is there anything else you suggest adding here?

src/core/sock/sock-redirect.cpp
alexbriskin: With the static library, there is no reason to use `getsockopt` with `SO_XLIO_GET_API`.

alexbriskin:
bot:retest

alexbriskin:
bot:retest

PR #91, opened 2024-02-02
issue: 3795922 Remove pbuf_split_64k() and refused_data

pasis:
## Description
Remove pbuf_split_64k() and refused_data. This code doesn't seem to be useful, but executes extra operations and touches extra cache line in fast TCP timer.

##### What
Remove pbuf_split_64k() and refused_data.

##### Why ?
Code cleanup, performance optimization.

## Change type
- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list


src/core/lwip/tcp_in.c
```diff
@@ -241,30 +216,18 @@ void L3_level_tcp_input(struct pbuf *p, struct tcp_pcb *pcb)
...
-                                pbuf_cat(in_data.recv_data, rest); /* undo splitting */
-                            }
-                            pcb->refused_data = in_data.recv_data;
-                            LWIP_DEBUGF(
-                                TCP_INPUT_DEBUG,
-                                ("tcp_input: keep incoming packet, because pcb is \"full\"\n"));
-                            break;
-                        } else {
-                            in_data.recv_data = rest;
+                            pcb->rcv_wnd += in_data.recv_data->tot_len;
```
AlexanderGrissik, line 229: Is there a scenario when upper layer refuses the data? What will happen if it does? will we corrupt TCP stream?

AlexanderGrissik (approved):

src/core/lwip/tcp_in.c
pasis: the only upper layer with RX support is TLS RX. Even if errors happen and TLS session is terminated, XLIO keeps the buffer and returns ERR_OK. The buffer is probably freed in destructor.

So, most likely scenario to fail the RX is when lwip triggers some error, passes the error to the recv() method and recv() returns it back. This is a confusing scheme.

galnoam:
@pasis, why this was not pushed as part of RM#3668182?
I see it was moved to close.

pasis:
Changed ticket number

pasis:
bot:retest

PR #96, opened 2024-02-18
issue: 3514044 Fix for 100% CPU load

iftahl:
Split poll_sn to poll_sn_rx and poll_sn_tx.
Required after splitting cq to cq_rx and cq_tx.

## Change type
- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list


pasis:
Do we need these `poll_sn`? In most cases we have to create fake variables on stack just to satisfy such API. If it's necessary in some specific cases, probably we can provide poll_sn only in those cases?

iftahl:
> Do we need these `poll_sn`? In most cases we have to create fake variables on stack just to satisfy such API. If it's necessary in some specific cases, probably we can provide poll_sn only in those cases?

maybe we can do something to optimize it, but im not really familiar with all the flows.
but when we split cq_mgr into 2 different classes, we needed to split also the poll_sn.

galnoam:
@AlexanderGrissik, please give it priority.
Thanks!

AlexanderGrissik (approved):

PR #98, opened 2024-02-21
issue: 3782717 Fairness

alexbriskin:
## Description
Improve fairness

## Change type
- [ ] Bugfix
- [X] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list

### DIFF
--- a/src/core/dev/ring_simple.cpp
+++ b/src/core/dev/ring_simple.cpp
@@ -256,7 +256,7 @@ void ring_simple::create_resources()
     }
     ring_logdbg("ring attributes: m_tx_num_wr = %d", m_tx_num_wr);
 
-#if 1
+#if 0
     m_tx_scheduler = std::make_unique<tx_fifo_scheduler>(*this, max_qp_wr);
 #else
     m_tx_scheduler = std::make_unique<tx_round_robin_scheduler>(*this, max_qp_wr);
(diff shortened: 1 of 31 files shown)


src/core/dev/ring_simple.cpp
```diff
@@ -241,7 +241,7 @@ void ring_simple::create_resources()
     VALGRIND_MAKE_MEM_DEFINED(m_p_tx_comp_event_channel, sizeof(struct ibv_comp_channel));
     // Check device capabilities for max QP work requests
     uint32_t max_qp_wr = ALIGN_WR_DOWN(m_p_ib_ctx->get_ibv_device_attr()->max_qp_wr);
-    m_tx_num_wr = safe_mce_sys().tx_num_wr;
+    m_sq_free_credits = m_tx_num_wr = safe_mce_sys().tx_num_wr;
```
pasis, line 244: `m_sq_free_credits` probably needs to be adjusted according to actual HW resources in case credit == WQEBB.

Some info for the context:
The previous implementation uses minimum between real SQ size and TX CQ size. We had to limit credits to the CQ size because of CQ overflow in corner cases:
Number of WQEs in flight (not WQEBBs) mustn't exceed CQ size to avoid the overflow. Since we don't track number of WQEs, we don't allow WQEBBs in flight more than CQ size.

PR #101, opened 2024-02-27
issue: 3788369 First drop of XLIO Socket / Storage API

pasis:
## Description
New zerocopy extra API with event driven approach and lightweight operations.
First drop. Provides API for static linkage.

##### What
New zerocopy extra API with event driven approach and lightweight operations.

##### Why ?
Performance improvement of certain workloads for applications that use XLIO explicitly.

## Change type
- [x] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list


src/core/sock/sockinfo.cpp
```diff
@@ -1968,53 +1962,48 @@ int sockinfo::modify_ratelimit(dst_entry *p_dst_entry, struct xlio_rate_limit_t
...
-        res_length = 1;
-        return m_p_rx_ring->get_rx_channel_fds(num_rx_channel_fds);
-    }
-
-    res_length = get_rings_num();
-    if (m_p_rings_fds) {
-        return m_p_rings_fds;
+    ring *tx_ring = m_p_connected_dst_entry ? m_p_connected_dst_entry->get_ring() : nullptr;
+    if (tx_ring) {
+        channel_fds = tx_ring->get_rx_channel_fds(num_rx_channel_fds);
```
AlexanderGrissik, line 1991: This will return m_p_rx_comp_event_channel. Do not we want the m_p_tx_comp_event_channel and use that as fd to poll ring TX?

pasis: This fd is used in `xlio_socketxtreme_poll()` to identify the ring. It doesn't matter whether this is RX or TX cq. However, taking always the same cq type simplifies the checks, because we can simply make an array of unique fds. With TX cq we would need to handle different fds of the same ring.

Besides SocketXtreme, `xlio_socket_get_rings_fsd()` is mainly used in tests to compare values across different sockets.

src/core/lwip/tcp.c
```diff
@@ -94,7 +94,6 @@ u16_t lwip_tcp_mss = CONST_TCP_MSS;
 u8_t enable_push_flag = 1;
 u8_t enable_ts_option = 0;
 u32_t lwip_tcp_snd_buf = 0;
-u32_t lwip_zc_tx_size = 0;
```
AlexanderGrissik, line 97: The field is removed but i can't find the usage to be removed.

pasis: Usage is removed in the following commit:
```
issue: 3788369 Pass iovec to tcp_write_express()
```
AI: Need to reorder commits and update the commit message that the option is unused.

src/core/Makefile.am
```diff
@@ -100,6 +100,7 @@ libxlio_la_SOURCES := \
 	event/delta_timer.cpp \
 	event/event_handler_manager.cpp \
 	event/event_handler_manager_local.cpp \
+	event/group.cpp \
```
AlexanderGrissik, line 103: I suggest to rename it to xlio_ring_group. Both the class and the file. Otherwise it suggests some general group code. But in fact this is a specific class for ring group

src/core/event/group.h
```diff
@@ -0,0 +1,79 @@
...
+    xlio_socket_rx_cb_t m_socket_rx_cb;
+
+private:
+    std::vector<ring *> m_rings;
+    event_handler_manager_local *m_event_handler;
+    tcp_timers_collection *m_tcp_timers;
+
+    std::vector<sockinfo_tcp *> m_dirty_sockets;
+
+    unsigned m_flags;
```
AlexanderGrissik, line 76: I suggest to call it something like m_ring_group_flags. It really makes it easier to do simple text search.

src/core/lwip/pbuf.h
```diff
@@ -70,6 +70,7 @@ typedef struct {
         void *mdesc;
         void *opaque;
     };
+    uintptr_t opaque_sq;
```
AlexanderGrissik, line 73: I suggest to rename to soemthing like opaque_tx_userdata or opaque_userdata or opaque_ctx

src/core/lwip/pbuf.h
```diff
@@ -117,7 +118,6 @@ typedef void (*pbuf_free_custom_fn)(struct pbuf *p);
 struct pbuf_custom {
```
AlexanderGrissik, line 118: Lets get rid of pbuf_custom now that it has only one member.

src/core/lwip/tcp_out.c
```diff
@@ -1537,6 +1552,11 @@ void tcp_split_segment(struct tcp_pcb *pcb, struct tcp_seg *seg, u32_t wnd)
             return;
         }
 
+        if (seg->p->desc.attr == PBUF_DESC_EXPRESS) {
```
AlexanderGrissik, line 1555: What about tcp_split_one_segment and tcp_split_rexmit?

src/core/event/group.cpp
```diff
@@ -0,0 +1,113 @@
...
+#define grp_logdbg   __log_dbg
+
+xlio_group::xlio_group(const struct xlio_group_attr *attr)
+    : m_socket_event_cb(attr->socket_event_cb)
+    , m_socket_comp_cb(attr->socket_comp_cb)
+    , m_socket_rx_cb(attr->socket_rx_cb)
+    , m_flags(attr->flags)
+{
+    m_rings.reserve(2);
+    m_event_handler = new event_handler_manager_local();
```
AlexanderGrissik, line 55: This better be unique_ptr and it will be destructed automatically on xlio_group destruction

src/core/event/group.cpp
```diff
@@ -0,0 +1,113 @@
...
+        sn = 0;
+        rng->poll_and_process_element_rx(&sn);
+    }
+    m_event_handler->do_tasks();
+}
+
+void xlio_group::add_dirty_socket(sockinfo_tcp *si)
+{
+    if (m_flags & XLIO_GROUP_FLAG_DIRTY) {
+        m_dirty_sockets.push_back(si);
```
AlexanderGrissik, line 81: I wonder if we want to reserve some initial space in the constructor in case XLIO_GROUP_FLAG_DIRTY is used....

src/core/event/group.cpp
```diff
@@ -0,0 +1,113 @@
...
+    }
+    m_dirty_sockets.clear();
+    // TODO Ring doorbell and request TX completion.
+}
+
+void xlio_group::add_ring(ring *rng)
+{
+    if (std::find(m_rings.begin(), m_rings.end(), rng) == std::end(m_rings)) {
+        grp_loginfo("New ring %p in group %p", rng, this);
+        rng->set_group(this);
```
AlexanderGrissik, line 98: I wonder if it worth checking if a ring already belongs to a group ....

src/core/Makefile.am
pasis: I wouldn't say this is a ring group. It's more socket group from user perspective and the ring collection is just implementation specific.
The idea of the group is to create a serialization entity or a context with a unique set of sockets. Operations with different groups can be executed in parallel and user needs to care about serialization only within a group.
Also, ring is an internal abstraction which shouldn't be exposed to user. Since we expose the group abstraction to user, I'd like to avoid mentioning the ring. If you think `xlio_group` is too generic, probably `xlio_socket_group` will be better?

src/core/lwip/tcp_out.c
pasis: They're removed in #98. I didn't touch the code not to make conflicts. Probably need to merge the removal to vNext outside the PR.

src/core/event/group.cpp
```diff
@@ -0,0 +1,113 @@
...
+{
+    delete m_tcp_timers;
+    delete m_event_handler;
+}
+
+void xlio_group::poll()
+{
+    for (ring *rng : m_rings) {
+        uint64_t sn;
+        rng->poll_and_process_element_tx(&sn);
```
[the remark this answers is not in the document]
pasis: In current implementation, xlio_group forces ring allocation logic which creates a single ring per ibdevice (per group). Without corner cases which lead to asymmetric routing, TX ring and RX ring will be the same. For expected proper usage of the API, keeping the rings in a single collection looks more efficient than keeping 2 separate collections. With recommendation to use a single ibdevice per group, there will be a single ring per group in the most efficient scenario.

src/core/event/group.cpp
pasis: I relied on the warmup phase here. Groups should be long lived, so after few iterations the vector should group big enough.
Plus, current implementation of the dirty sockets collection is very limited and I need to rework it somehow to make it interlive with socket level flush and flushing send().

src/core/event/group.cpp
pasis: Probably as sanity check or assert. If we see a ring here, that means the socket uses special ring allocation logic with the key equals to the group pointer. Therefore, in GA version this shouldn't be possible.

Current implementation can lead to a situation when a group is destroyed without destroying the sockets and then new group is allocated with the same address - the old ring will be reused for the new group. Even such a corner case should work functionally. GA version will terminate (reset) the remaining sockets on a group destruction, so the ring will be released.

There is AI for GA version to cover the above. Will add a sanity check for current version that a group is not set.

src/core/lwip/tcp.c
pasis: Done. Usage removed in the right commit now.

src/core/lwip/pbuf.h
pasis: Done.

src/core/event/group.cpp
pasis: Done. Added warning in case of a group is already set.

src/core/event/group.cpp
pasis: Done.

src/core/lwip/pbuf.h
pasis: Done. Removed the field completely, because current implementation calls the callback from sockinfo_tcp context, so it has access to the value from the socket object.

pasis:
@AlexanderGrissik, updates are in the new commits (that follow `issue: 3788369 Storage API`)

src/core/xlio.h
```diff
@@ -382,6 +384,225 @@ int xlio_socketxtreme_ref_buff(struct xlio_buff_t *buff);
...
+/*
+ * Memory callback.
+ *
+ * XLIO calls the callback each time XLIO allocates a memory region which can be used for RX
+ * buffers. User can use this information to prepare the memory for some logic in the future.
+ * Zerocopy RX interface provides pointers to such memory.
+ *
+ * Current XLIO implementation does a single allocation for buffers.
+ */
+typedef void (*xlio_memory_cb_t)(void *addr, size_t len, size_t hugepage_size);
```
benlwalker, line 419: Does xlio still support XLIO_IOCTL_USER_ALLOC? We need to allocate the memory outside of xlio, rather than just be notified of what occurred.

pasis: At the moment it's still supported. However, I wanted to remove it eventually.

Since you need external allocation, we need to discuss the reason and possible solution, because current implementation of `XLIO_IOCTL_USER_ALLOC` is inconvenient. For example, we can simplify the external allocator (remove differentiation between TX and RX specific allocation) and provide it to `xlio_init_ex()` instead of inconvenient `ioctl()` call).

src/core/Makefile.am
AlexanderGrissik: xlio_sockets_group sounds right!

src/core/xlio.h
```diff
@@ -382,6 +384,225 @@ int xlio_socketxtreme_ref_buff(struct xlio_buff_t *buff);
...
+
+struct xlio_init_attr {
+    unsigned flags;
```
AlexanderGrissik, line 422: Consider sing something that can easily be 'grep'd without huge duplicates in the code, like init_flags or anything that can be distiguished in simple text search.

src/core/xlio.h
```diff
@@ -382,6 +384,225 @@ int xlio_socketxtreme_ref_buff(struct xlio_buff_t *buff);
...
+ *
+ * May be called from the following contexts:
+ *  - xlio_group_poll() - likely
+ *  - xlio_socket_send() - can happen only if data is flushed
+ *  - xlio_socket_flush() / xlio_group_flush()
+ *  - xlio_socket_destroy()
+ *
+ * In the callback context, send operation is allowed unless the socket is under destruction.
+ */
+typedef void (*xlio_socket_comp_cb_t)(xlio_socket_t, uintptr_t userdata_sq, uintptr_t userdata_op);
```
AlexanderGrissik, line 469: Should we provide a way for a user to know if the socket is under destruction? or is it enough to let the user track CLOSE/TERMINATED events?

src/core/xlio.h
```diff
@@ -382,6 +384,225 @@ int xlio_socketxtreme_ref_buff(struct xlio_buff_t *buff);
...
+ *    optimize the HW objects utilization. However, maintaining extra groups can have an overhead.
+ */
+
+/* Sockets and rings will be protected with locks regardless of XLIO configuration. */
+#define XLIO_GROUP_FLAG_SAFE 0x1
+/* Group will keep dirty sockets to be flushed with xlio_group_flush(). */
+#define XLIO_GROUP_FLAG_DIRTY 0x2
+
+struct xlio_group_attr {
+    unsigned flags;
```
AlexanderGrissik, line 504: Same as above

src/core/xlio.h
```diff
@@ -382,6 +384,225 @@ int xlio_socketxtreme_ref_buff(struct xlio_buff_t *buff);
...
+ * Example of the possible events is zerocopy completions which can arrive from the
+ * xlio_socket_destroy() context or xlio_group_poll() context.
+ *
+ * Limitations:
+ *  - Only outgoing connections are supported
+ *  - Bonding is not supported
+ */
+
+struct xlio_socket_attr {
+    unsigned flags;
```
AlexanderGrissik, line 532: same as above

src/core/xlio.h
```diff
@@ -382,6 +384,225 @@ int xlio_socketxtreme_ref_buff(struct xlio_buff_t *buff);
...
+ *    it's better to avoid using them both.
+ */
+
+/* Flush socket after queueing the data. */
+#define XLIO_SOCKET_SEND_FLAG_FLUSH 0x1
+/* Copy user data to the internal buffers instead of taking ownership. */
+#define XLIO_SOCKET_SEND_FLAG_INLINE 0x2
+
+struct xlio_socket_send_attr {
+    unsigned flags;
```
AlexanderGrissik, line 578: same as above

src/core/lwip/tcp_out.c
```diff
@@ -748,6 +749,12 @@ err_t tcp_write_express(struct tcp_pcb *pcb, const struct iovec *iov, u32_t iovc
     }
     last_seglen = last ? last->len : 0;
 
+    if (desc->attr == PBUF_DESC_EXPRESS) {
+        /* Keep opaque value only in the right most pbuf for each send operation. */
+        opaque = desc->opaque;
+        desc->opaque = NULL;
```
AlexanderGrissik, line 755: Consider adding here a comment why we null the desc->opaque and then restore it.

src/core/sock/sock-extra.cpp
```diff
@@ -373,3 +375,189 @@ struct xlio_api_t *extra_api(void)
...
+}
+
+extern "C" int xlio_socket_connect(xlio_socket_t sock, const struct sockaddr *to, socklen_t tolen)
+{
+    sockinfo_tcp *si = reinterpret_cast<sockinfo_tcp *>(sock);
+
+    int rc = XLIO_CALL(connect, si->get_fd(), to, tolen);
+    rc = (rc == -1 && (errno == EINPROGRESS || errno == EAGAIN)) ? 0 : rc;
+    if (rc == 0) {
+        si->add_tx_ring_to_group();
```
AlexanderGrissik, line 490: What happens with non-blocking socket? when will it be added to the group?

src/core/sock/sockinfo_tcp.h
```diff
@@ -304,6 +353,7 @@ class sockinfo_tcp : public sockinfo, public timer_handler {
         return m_p_connected_dst_entry ? m_p_connected_dst_entry->get_ring() : nullptr;
     }
 
+    void rx_add_ring_cb(ring *p_ring);
```
AlexanderGrissik, line 356: Please specify 'override' so it's clear for the reader that this is a virtual method.

src/core/sock/sockinfo_tcp.cpp
```diff
@@ -385,6 +396,127 @@ sockinfo_tcp::sockinfo_tcp(int fd, int domain)
...
+    }
+
+    tcp_recved(pcb, p->tot_len);
+
+    if (conn->m_p_group->m_socket_rx_cb) {
+        struct pbuf *ptmp = p;
+        while (ptmp) {
+            /* TODO Pass mem_buf_desc_t field intead of pbuf itself as xlio_buf */
+            conn->m_p_group->m_socket_rx_cb(reinterpret_cast<xlio_socket_t>(conn),
+                                            conn->m_xlio_socket_userdata, ptmp->payload, ptmp->len,
```
AlexanderGrissik, line 475: Need to verify that in case of GRO/out-of-order payload and len are always correct for all parts individually.

src/core/xlio.h
pasis: I would leave this to user to track timeline after `xlio_socket_destroy()` and before `XLIO_SOCKET_EVENT_TERMINATED`. Adding such information would unnecessary complicate the API. Better rely on the `xlio_socket_send()` failure, because inner check for `is_rts()` will fail.

src/core/sock/sockinfo_tcp.h
pasis: I did it first, but then clang start complaining about all the overridden methods in the header. Should I put `override` to the all methods in this header as part of current PR?

src/core/sock/sockinfo_tcp.cpp
pasis: The pbuf chain must be valid (tot_len, len, payload). Otherwise, this would be a bug and RX operations would lead to unexpected behavior. `rx_lwip_process_chained_pbuf()` can be a reference that we rely on the chain validity on each RX operation.

src/core/event/group.h
```diff
@@ -57,9 +58,9 @@ class xlio_group {
     void add_ring(ring *);
     void del_ring(ring *);
 
-    inline unsigned get_flags() { return m_flags; }
-    inline event_handler_manager_local *get_event_handler() { return m_event_handler; }
-    inline tcp_timers_collection *get_tcp_timers() { return m_tcp_timers; }
+    inline unsigned get_flags() const { return m_flags; }
```
AlexanderGrissik, line 61: AFAIK inline is not needed when implemented in place. This is implicit.

src/core/sock/sockinfo_tcp.h
AlexanderGrissik: Not a must. I will do it as part of cache work i hope.

src/core/sock/sockinfo_tcp.cpp
```diff
@@ -6061,13 +6234,63 @@ int sockinfo_tcp::tcp_tx_express(const struct iovec *iov, unsigned iov_len, uint
...
+            // XXX tcp_write() can also fail due to queuelen limit, but this is unlikely.
+            m_conn_state = TCP_CONN_ERROR;
+            m_error_status = ENOMEM;
+            return tcp_tx_handle_errno_and_unlock(ENOMEM);
+        }
+    }
+    if (!(flags & XLIO_EXPRESS_MSG_MORE)) {
+        m_b_xlio_socket_dirty = false;
+        tcp_output(&m_pcb);
+    } else if (m_p_group && !m_b_xlio_socket_dirty) {
```
AlexanderGrissik, line 6271: Should we check here if dirty handling is requested?

src/core/sock/sockinfo_tcp.cpp
```diff
@@ -6061,13 +6234,63 @@ int sockinfo_tcp::tcp_tx_express(const struct iovec *iov, unsigned iov_len, uint
     }
     if (!(flags & XLIO_EXPRESS_MSG_MORE)) {
         tcp_output(&m_pcb);
+        m_b_xlio_socket_dirty = false;
+    } else if (m_p_group && !m_b_xlio_socket_dirty) {
```
AlexanderGrissik, line 6233: Should we check if dirty handling is requested?

src/core/sock/sockinfo.cpp
AlexanderGrissik: Lets then add a comment here that in future others don't think it is a bug.

src/core/Makefile.am
pasis: will rename to `xlio_polling_group`

src/core/event/group.h
pasis: will remove

src/core/event/group.h
pasis: Done, renamed to m_group_flags

src/core/Makefile.am
pasis: Done.

src/core/sock/sock-extra.cpp
pasis: Done, errno is restored for EINPROGRESS case (as it's a successful call).

src/core/event/group.h
pasis: Done.

src/core/sock/sockinfo_tcp.cpp
pasis: Currently it's checked inside the poll_group method, because this is the group configuration whether to track dirty sockets or not.

src/core/sock/sockinfo_tcp.cpp
pasis: Currently it's checked inside the poll_group method, because this is the group configuration whether to track dirty sockets or not.

src/core/xlio.h
pasis: I will leave it as is because:
1. The attribute structure describes specific object/subsystem and it's unambiguous as result
2. Simple field "flags" follows the well known APIs (like POSIX), so this also follows least surprise principle
3. This is public API and shouldn't go deep into XLIO implementation, therefore, the common word "flags" shouldn't make big impact on the text grep.

src/core/xlio.h
pasis: Please see comment above

src/core/xlio.h
pasis: Please see comment above

src/core/xlio.h
pasis: Please see comment above

src/core/sock/sockinfo.cpp
pasis: Done.

AlexanderGrissik (approved):

galnoam (approved):

PR #109, opened 2024-03-12
issue: 3788369 Second drop of Storage / XLIO socket API

pasis:
## Description
XLIO socket API is a performance oriented API.

Note, rebased on top of #115
Begins at `issue: 3788369 Remove m_required_send_block introduced by NVME offload`

##### What
Productization of the 1st drop.

##### Why ?
Productization of the 1st drop.

## Change type
- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list

### DIFF
--- a/src/core/dev/ring_simple.cpp
+++ b/src/core/dev/ring_simple.cpp
@@ -90,6 +90,13 @@ ring_simple::ring_simple(int if_index, ring *parent, ring_type_t type, bool use_
     , m_gro_mgr(safe_mce_sys().gro_streams_max, MAX_GRO_BUFS)
 {
     net_device_val *p_ndev = g_p_net_device_table_mgr->get_net_device_val(m_parent->get_if_index());
+    BULLSEYE_EXCLUDE_BLOCK_START
+    if (!p_ndev) {
+        // Coverity warning suppression
+        throw_xlio_exception("Cannot find netdev for a ring");
+    }
+    BULLSEYE_EXCLUDE_BLOCK_END
+
     const slave_data_t *p_slave = p_ndev->get_slave(get_if_index());
 
     ring_logdbg("new ring_simple()");
@@ -209,10 +216,18 @@ ring_simple::~ring_simple()
 void ring_simple::create_resources()
 {
     net_device_val *p_ndev = g_p_net_device_table_mgr->get_net_device_val(m_parent->get_if_index());
+    BULLSEYE_EXCLUDE_BLOCK_START
+    if (!p_ndev) {
+        // Coverity warning suppression
+        throw_xlio_exception("Cannot find netdev for a ring");
+    }
+    BULLSEYE_EXCLUDE_BLOCK_END
+
     const slave_data_t *p_slave = p_ndev->get_slave(get_if_index());
 
     save_l2_address(p_slave->p_L2_addr);
     m_p_tx_comp_event_channel = ibv_create_comp_channel(m_p_ib_ctx->get_ibv_context());
+    BULLSEYE_EXCLUDE_BLOCK_START
     if (!m_p_tx_comp_event_channel) {
         VLOG_PRINTF_INFO_ONCE_THEN_ALWAYS(
             VLOG_ERROR, VLOG_DEBUG,
@@ -319,9 +334,7 @@ void ring_simple::create_resources()
 #endif
     ring_logdbg("ring attributes: m_flow_tag_enabled = %d", m_flow_tag_enabled);
 
-    m_p_rx_comp_event_channel = ibv_create_comp_channel(
-        m_p_ib_ctx->get_ibv_context()); // ODED TODO: Adjust the ibv_context to be the exact one in
-                                        // case of different devices
+    m_p_rx_comp_event_channel = ibv_create_comp_channel(m_p_ib_ctx->get_ibv_context());
     BULLSEYE_EXCLUDE_BLOCK_START
     if (!m_p_rx_comp_event_channel) {
         VLOG_PRINTF_INFO_ONCE_THEN_ALWAYS(
--- a/src/core/event/poll_group.cpp
+++ b/src/core/event/poll_group.cpp
@@ -86,6 +86,13 @@ poll_group::~poll_group()
     }
     s_poll_groups_lock.unlock();
 
+    while (!m_sockets_list.empty()) {
+        sockinfo_tcp *si = dynamic_cast<sockinfo_tcp *>(m_sockets_list.front());
+        if (likely(si)) {
+            close_socket(si, true);
+        }
+    }
+
     grp_logdbg("Polling group %p destroyed", this);
 }
 
@@ -149,4 +156,20 @@ void poll_group::del_ring(ring *rng)
         grp_logdbg("Removed ring %p from group %p", rng, this);
         m_rings.erase(iter);
     }
+}
+
+void poll_group::add_socket(sockinfo_tcp *si)
+{
+    m_sockets_list.push_back(si);
+}
+
+void poll_group::close_socket(sockinfo_tcp *si, bool force /*=false*/)
+{
+    m_sockets_list.erase(si);
+
+    bool closed = si->prepare_to_close(force);
+    if (closed) {
+        si->clean_socket_obj();
+    }
+    // TODO If not closed, the socket will be destroyed after the last completion notification.
 }
--- a/src/core/event/poll_group.cpp
+++ b/src/core/event/poll_group.cpp
@@ -93,6 +93,8 @@ poll_group::~poll_group()
         }
     }
 
+    // TODO Release the rings. Current implementation destroys the rings in the library destructor.
+
     grp_logdbg("Polling group %p destroyed", this);
 }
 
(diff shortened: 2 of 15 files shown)


src/core/sock/sockinfo_tcp.cpp
```diff
@@ -1112,6 +1102,10 @@ ssize_t sockinfo_tcp::tcp_tx(xlio_tx_call_attr_t &tx_arg)
                 return tcp_tx_handle_sndbuf_unavailable(total_tx, is_dummy, is_non_file_zerocopy,
                                                         errno_tmp);
             }
+            if (tx_arg.xlio_flags & TX_FLAG_NO_PARTIAL_WRITE) {
+                // TODO Fix type overflow if user sends over UINT32_MAX bytes
```
AlexanderGrissik, line 1101: Would it be acceptable to return error for now if uint32 exceeded?

src/core/event/poll_group.cpp
```diff
@@ -61,10 +69,39 @@ poll_group::poll_group(const struct xlio_poll_group_attr *attr)
...
+
+    grp_logdbg("Polling group %p destroyed", this);
+}
+
+/*static*/
+void poll_group::destroy_all_groups()
+{
+    poll_group *grp;
+
+    do {
```
AlexanderGrissik, line 106: Can just copy the list to another list with one line and then iterate the new list with for (xx:xx). It doesn't have to be efficient anyway and will very simplify the code.

src/core/event/poll_group.cpp
```diff
@@ -154,3 +161,19 @@ void poll_group::del_ring(ring *rng)
...
+
+void poll_group::close_socket(sockinfo_tcp *si, bool force /*=false*/)
+{
+    m_sockets_list.erase(si);
+
+    bool closed = si->prepare_to_close(force);
+    if (closed) {
+        si->clean_socket_obj();
+    }
+    // TODO If not closed, the socket will be destroyed after the last completion notification.
```
AlexanderGrissik, line 179: Should it be something that needs still be handled? Will the socket still live in timers pending to remove?

src/core/dev/ring_simple.cpp
```diff
@@ -90,6 +90,12 @@ ring_simple::ring_simple(int if_index, ring *parent, ring_type_t type, bool use_
     , m_gro_mgr(safe_mce_sys().gro_streams_max, MAX_GRO_BUFS)
 {
     net_device_val *p_ndev = g_p_net_device_table_mgr->get_net_device_val(m_parent->get_if_index());
+    BULLSEYE_EXCLUDE_BLOCK_START
+    if (!p_ndev) {
+        throw_xlio_exception("Cannot find netdev for a ring");
```
AlexanderGrissik, line 95: Is this situation possible? Should we have a reference of a netdev if a ring is related to it?

src/core/event/poll_group.cpp
```diff
@@ -138,27 +140,26 @@ void poll_group::flush()
...
+        /*
+         * Take reference to the ring. This avoids a race between socket destruction and buffer
+         * return to the group. Socket destruction can lead to the ring destruction. But user
+         * may return a buffer outside of the socket lifecycle.
+         * This also avoids extra ring destruction in a scenario when application closes all
+         * the sockets multiple times in runtime.
+         */
+        net_device_val *nd = g_p_net_device_table_mgr->get_net_device_val(rng->get_if_index());
+        if (nd) {
+            ring *reserved = nd->reserve_ring(attr);
```
AlexanderGrissik, line 158: How net_dev will destroy rings if we don't dec ref count on group destruction?

src/core/event/poll_group.h
```diff
@@ -57,8 +58,7 @@ class poll_group {
     void add_dirty_socket(sockinfo_tcp *si);
     void flush();
 
-    void add_ring(ring *);
-    void del_ring(ring *);
```
AlexanderGrissik, line 59: btw, when scket is deleted and it's ring is no longer needed in the group, it still will remain in the group....

src/core/event/poll_group.cpp
pasis: done.

src/core/event/poll_group.cpp
pasis: This is TODO for current PR. We need to keep the socket object until all the TX completions are handled. This is different then pending to remove, because in current case, we are ready to destroy the socket object.

src/core/dev/ring_simple.cpp
pasis: This situation doesn't look possible, but coverity generates a warning. So this is just a suppression.

src/core/event/poll_group.cpp
pasis: Added ref release to the destructor. Previously the rings were destroyed in the net_device_val destructor during XLIO termination.

Done.

src/core/event/poll_group.h
pasis: Yes, this is intentional to avoid ring thrashing in the scenario when application removes all the sockets and then creates new ones frequently. Ring destruction is too expensive now.

Also one of the main reasons to keep rings is efficient RX buffer release which can be performed after the socket destruction. Without keeping the ring, we would need to search for the ring in the group collection and return the buffer to the global buffer pool in corner cases. This would have a cost for the extra lookup.

src/core/sock/sockinfo_tcp.cpp
pasis: as last resort yes. Other option is to make send buffer 64bit which will increase socket size by 4 bytes.

(not done yet)

src/core/sock/sock-extra.cpp
```diff
@@ -549,7 +549,11 @@ static void xlio_buf_free(struct xlio_buf *buf)
     mem_buf_desc_t *desc = reinterpret_cast<mem_buf_desc_t *>(buf);
     ring_slave *rng = desc->p_desc_owner;
 
-    (void)rng->reclaim_recv_single_buffer(desc);
+    desc->p_next_desc = nullptr;
+    bool ret = rng->reclaim_recv_buffers(desc);
```
AlexanderGrissik, line 560: Not related to this change. What is the idea of keeping the mentioned list of buffers (m_rx_buffs_rdy_for_free_head) ? Should we remove completely?

src/core/event/poll_group.cpp
```diff
@@ -182,7 +182,15 @@ void poll_group::close_socket(sockinfo_tcp *si, bool force /*=false*/)
...
     if (closed) {
+        /*
+         * Current implementation forces TCP reset, so the socket is expected to be closable.
+         * Do a polling iteration to increase the chance that all the relevant WQEs are completed
+         * and XLIO emitted all the TX completion before the XLIO_SOCKET_EVENT_TERMINATED event.
+         *
+         * TODO Implement more reliable mechanism of deferred socket destruction if there are
+         * not completed TX operations.
+         */
+        poll();
```
AlexanderGrissik, line 193: AFAIU it will not affect time-wait sockets. So while a socket in time-wait it's TX completions can be triggered?

src/core/sock/sock-extra.cpp
```diff
@@ -406,6 +406,9 @@ extern "C" int xlio_init_ex(const struct xlio_init_attr *attr)
         setenv(SYS_VAR_PROGRESS_ENGINE_INTERVAL, "0", 1);
     }
 
+    // Read the updated parameters. A global object could trigger the reading earlier.
+    safe_mce_sys().get_env_params();
```
AlexanderGrissik, line 410: Maybe a note to XLIO Gaps: We need to remove any possibly unordered initialization in global constructors and do initialization in a controlled manner..

src/core/dev/ring_simple.cpp
AlexanderGrissik: Lets add a comment that its done for Coverity, so we don't remove it or think that we obviously have this situation.

src/core/event/poll_group.h
AlexanderGrissik: Looks like an important comment to add :)

src/core/util/xlio_list.h
```diff
@@ -241,23 +144,27 @@ template <class T, size_t offset(void)> class xlio_list_t {
...
-/* clang analyzer reports:
- * Use of memory after it is freed
- * This issue comes from ~chunk_list_t()
- * Root cause is unknown.
- * TODO: Fix based on root cause instead of supressing
- */
-#ifndef __clang_analyzer__
-        return ((list_node<T, offset> *)m_list.head.prev)->obj_ptr;
-#endif
+        return ((list_node<T, offset> *)m_list.head.prev)->obj_ptr();
```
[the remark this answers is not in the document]
pasis: Moved to other PR

src/core/event/poll_group.cpp
pasis: To reduce the chance of this scenario, polling group method does a polling iteration to collect remaining TX completions. Usually a socket is closed in a state where there are no pending data. But even if a socket is closed in the middle of TCP stream, we force TCP reset, so the only problematic scenario is when there WQEs in SQ that are not completed until to polling group poll() inside the close method. This scenario will be addressed in future version.

src/core/dev/ring_simple.cpp
pasis: Done.

src/core/event/poll_group.h
pasis: There is already a comment in the `poll_group::add_ring()` implementation

src/core/sock/sock-extra.cpp
pasis: The list reminds me other similar lists in sockets and other places. The idea was likely to accumulate single objects first and do the handling in a batch. However, this is not cache efficient.
I agree that we should handle single object without accumulation, but it need more deeper analysis of the impact.

src/core/event/poll_group.cpp
pasis: Current implementation doesn't support time wait sockets. Instead, XLIO always forces TCP reset for the xlio_socket_t objetcs. On TCP reset, all the queues (unsent, unacked) are cleared immediately and socket can be destroyed without a delay.

In some future implementation, time-wait state or async TCP FIN handshake will be transparent for users, because API allows TX completions after xlio_socket_destroy() until async event TERMINATED arrives.

Also we cannot wait until socket destructor for TIME WAIT sockets, because it can take 1 minute to generate TERMINATED event. So we will need either speedup TIME WAIT destruction according to RFCs or generate TERMINATED event when socket transits to the TIME WAIT state. However, with the latter solution we will get race for RFC reuse issue.

src/core/sock/sock-extra.cpp
pasis: will add a note in the document

pasis:
bot:retest

PR #110, opened 2024-03-13
issue: 3777348 Move socket statistics to a global storage

AlexanderGrissik:
## Description
This PR is based on PR 106 (issue: 3777348 Simplifying sockinfo inheritance)
This PR starts from commit (issue: 3777348 Moving sock stats outside the socket)

##### What
Socket statistics consume a big chunk of memory in each socket separately.
Even in case statistics are not needed these stats take memory and are accesses and updated.
This PR moves statistics to a global vector and sockets ask for stats object.
Sockets will receive stats objects up to configured max stats. Other sockets will not have stats. 
This allows:
1. Saving memory when less stats are needed then amount of sockets (Like 1M sockets)
2. Disable stats and save CPU cycles and cache access.

##### Why ?
Performance optimization

## Change type
- [ ] Bugfix
- [X] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list


src/core/sock/sock_stats.cpp
```diff
@@ -0,0 +1,81 @@
...
+    std::lock_guard<decltype(_stats_lock)> lock(_stats_lock);
+
+    _socket_stats_vec.resize(max_stats);
+    for (size_t idx = 1; idx < _socket_stats_vec.size(); ++idx) {
+        _socket_stats_vec[idx - 1U]._next_stat = &_socket_stats_vec[idx];
+    }
+
+    _socket_stats_vec[_socket_stats_vec.size() - 1U]._next_stat = nullptr;
+    _socket_stats_first = &_socket_stats_vec[0];
+}
```
pasis, line 81: Please add end-of-line and the end of the file.

src/core/sock/sock_stats.cpp
```diff
@@ -0,0 +1,81 @@
...
+    return s_sock_stats.get();
+}
+
+socket_stats_t *sock_stats::get_stats_obj()
+{
+    if (!_socket_stats_first)
+        return nullptr;
+
+    std::lock_guard<decltype(_stats_lock)> lock(_stats_lock);
+    auto *stat = _socket_stats_first;
```
pasis, line 54: Please replace auto with `socket_stats_t` - this will simplify code reading.

src/core/sock/sock_stats.cpp
```diff
@@ -0,0 +1,81 @@
...
+    if (unlikely(!s_sock_stats)) {
+        s_sock_stats.reset(new sock_stats());
+    }
+
+    return s_sock_stats.get();
+}
+
+socket_stats_t *sock_stats::get_stats_obj()
+{
+    if (!_socket_stats_first)
```
pasis, line 50: According to the coding style, all the control blocks must be with brackets {}. Looks like CI failed to catch this.

pasis, line 50: Is it safe to access `_socket_stats_first` without a lock? Can it happen that for a single object in the list, multiple threads will pass the condition and then in the critical section the 2nd thread will dereference NULL unconditionally?

src/core/sock/sock_stats.h
```diff
@@ -0,0 +1,62 @@
...
+    socket_stats_t *get_stats_obj();
+    void return_stats_obj(socket_stats_t *stats);
+
+private:
+    static std::unique_ptr<sock_stats> s_sock_stats;
+
+    sock_stats() {}
+
+    std::mutex _stats_lock;
+    socket_stats_t *_socket_stats_first = nullptr;
```
pasis, line 58: [suggestion] Replace `_first` suffix with `_list`.

src/core/util/xlio_stats.h
```diff
@@ -225,44 +225,51 @@ typedef struct socket_listen_counters {
...
-    pid_t threadid_last_rx;
-    pid_t threadid_last_tx;
-    uint32_t n_rx_ready_pkt_count;
-    uint32_t n_rx_ready_byte_limit;
+struct socket_stats_t {
+    // Data Path
+
+    union {
+        uint64_t n_tx_ready_byte_count;
+        socket_stats_t *_next_stat;
```
pasis, line 233: AFAIU, `_next_stat` is used only to keep a stat object in the global free list. If yes, I'd suggest to move it to the cold zone (for example, to the end on the structure). Extra 8 cold bytes won't make an impact, but this will improve readability.

src/core/sock/sockinfo.h
```diff
@@ -328,6 +329,8 @@ class sockinfo {
     inline bool set_flow_tag(uint32_t flow_tag_id);
     inline void sock_pop_descs_rx_ready(descq_t *cache);
 
+    bool has_stats() const { return m_has_stats; }
```
pasis, line 332: [question] Have you considered --disable-socket-stats option will converts `has_stats()` to false in compile time? This will optimize out a lot of code.

AlexanderGrissik: Yes. i considered it. I even tried to comment out all the stats in DP and didn't see any change in IOPs. I decided to keep it this way to avoid another compile time flag.

src/core/sock/sock_stats.cpp
AlexanderGrissik: It shouldn't be an issue because code expects to get nullptr. But i guess it would be safer/correct to put the lock from the beggining.

src/core/sock/sock_stats.cpp
```diff
@@ -0,0 +1,81 @@
...
+ */
+
+#include "sock_stats.h"
+
+std::unique_ptr<sock_stats> sock_stats::s_sock_stats;
+thread_local socket_stats_t sock_stats::t_dummy_stats;
+
+sock_stats *sock_stats::get_sock_stats()
+{
+    if (unlikely(!s_sock_stats)) {
```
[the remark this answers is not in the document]
AlexanderGrissik: Done

src/core/sock/sock_stats.cpp
AlexanderGrissik: Done

src/core/sock/sock_stats.cpp
AlexanderGrissik: Done

src/core/sock/sock_stats.cpp
AlexanderGrissik: Done.

AlexanderGrissik: Done

src/core/sock/sock_stats.h
AlexanderGrissik: Done

src/core/util/xlio_stats.h
AlexanderGrissik: Done

pasis (approved):

PR #115, opened 2024-03-18
issue: 3788369 Update Storage / XLIO Socket public API

pasis:
## Description
Decouple public API related changes from the 2nd drop of the feature. This simplifies review and verification.

##### What
Update the public update after the 1st drop.

##### Why ?
Make public API more stable to simplify integration.

## Change type
- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list


src/core/xlio_types.h
```diff
@@ -0,0 +1,465 @@
...
+/*
+ * XLIO Socket API attribute structures
+ */
+
+struct xlio_init_attr {
+    unsigned flags;
+    xlio_memory_cb_t memory_cb;
+
+    /* Optional external user allocator for XLIO buffers. */
+    void *(*memory_alloc)(size_t);
```
AlexanderGrissik, line 430: I believe this was a request to add external allocator. And it will  be used to allocate both TX and RX buffers. Is this OK that the same allocator used for both?

AlexanderGrissik (approved):

AlexanderGrissik:
bot:retest

src/core/xlio_types.h
pasis: This is done intentionally to simplify the default logic for user and XLIO. If differentiation between RX and TX memory becomes critical for some reason we will add a flag to `xlio_init_attr` to control that.

PR #121, opened 2024-03-26
issue: 3795997 Allow split segment with unacked q

iftahl:
When send window is not big enough for the required TCP segment to send, we may split the segment so it will fir into the window. Before this change - We didn't split the segment in the case we have unacked segments. The motivation was that we anticiapte to get ACK on the inflight segments, which will trigger the next send operation.
This flow count on RTT for receiving ACKs, which may be delayed depending on the remote side. When RTT is long - we would block sending although TCP send window allows it.

The change is to split TCP segments although we have unacked data, in case the send window is big enough (mss).

## Change type
- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list

### DIFF
--- a/src/core/lwip/tcp_out.c
+++ b/src/core/lwip/tcp_out.c
@@ -1813,7 +1813,8 @@ err_t tcp_output(struct tcp_pcb *pcb)
         }
 
         /* Split the segment in case of a small window */
-        if ((NULL == pcb->unacked) && (wnd) && ((seg->len + seg->seqno - pcb->lastack) > wnd)) {
+        if (wnd && ((NULL == pcb->unacked) || (wnd >= pcb->mss)) &&
+            ((seg->len + seg->seqno - pcb->lastack) > wnd)) {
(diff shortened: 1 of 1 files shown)


src/core/lwip/tcp_out.c
```diff
@@ -1813,7 +1813,8 @@ err_t tcp_output(struct tcp_pcb *pcb)
         }
 
         /* Split the segment in case of a small window */
-        if ((NULL == pcb->unacked) && (wnd) && ((seg->len + seg->seqno - pcb->lastack) > wnd)) {
+        if (wnd && ((NULL == pcb->unacked) || (wnd > (pcb->mss / 2))) &&
```
AlexanderGrissik, line 1816: Why specifically mss/2?

pasis (approved):
Approved after changing condition to `wnd >= mss` instead of `> mss/2`

src/core/lwip/tcp_out.c
iftahl: switched to >=mss

AlexanderGrissik (approved):

PR #123, opened 2024-04-02
issue: 3818038 Remove BlueFlame doorbell method

pasis:
## Description
rdma-core limits number of UARs per context to 16 by default. After
creating 16 QPs, XLIO receives duplicates of blueflame registers for
each subsequent QP. As results, blueflame doorbell method can write WQEs
concurrently without serialization and this leads to data corruption.

BlueFlame can make impact on throughput since copy to the blueflame
register is expensive. It can improve latency in some low latency
scenarios, however, XLIO targets high traffic/PPS rates.
Removing blueflame method slightly improves performance in some
scenarios.

BlueFlame can be returned in the future to improve low-latency
scenarios, however, it will need some rework to avoid the data
corruption.

##### What
Remove BlueFlame doorbell method

##### Why ?
Fix data corruption for multi-threaded applications with over 16 QPs. Minor throughput/RPS improvement in some scenarios.

## Change type
- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list

### DIFF
--- a/src/core/dev/hw_queue_tx.cpp
+++ b/src/core/dev/hw_queue_tx.cpp
@@ -140,11 +140,6 @@ hw_queue_tx::hw_queue_tx(ring_simple *ring, const slave_data_t *slave, const uin
     // Check device capabilities for dummy send support
     m_hw_dummy_send_support = xlio_is_nop_supported(m_p_ib_ctx_handler->get_ibv_device_attr());
 
-    m_db_method =
-        (is_bf((slave->p_ib_ctx)->get_ibv_context()) ? MLX5_DB_METHOD_BF : MLX5_DB_METHOD_DB);
-
-    hwqtx_logdbg("m_db_method=%d", m_db_method);
-
     if (configure(slave)) {
         throw_xlio_exception("Failed to configure");
     }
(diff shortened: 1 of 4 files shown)


src/core/dev/hw_queue_tx.cpp
```diff
@@ -498,28 +493,19 @@ void hw_queue_tx::init_queue()
...
-        } else {
-#if defined(DEFINED_IBV_DM)
-            VLOG_PRINTF_ONCE_THEN_DEBUG(
-                VLOG_WARNING,
-                "Device Memory functionality is not used on devices w/o Blue Flame support\n");
-#endif /* DEFINED_IBV_DM */
-        }
+    if (m_p_ib_ctx_handler->get_on_device_memory_size() > 0 && is_bf(m_p_ib_ctx_handler->get_ibv_context())) {
+        m_dm_enabled =
+            m_dm_mgr.allocate_resources(m_p_ib_ctx_handler, m_p_ring->m_p_ring_stat.get());
```
AlexanderGrissik, line 508: Do we know if device memory can be used when BF is not?

AlexanderGrissik (approved):

src/core/dev/hw_queue_tx.cpp
pasis: From the XLIO perspective, we allocate device memory explicitly with ibv_alloc_dm(). Therefore an issue similar to BF duplicates is unlikely.
Also, device memory is independent from the doorbell method. According to the comment, it's disabled for devices without blueflame due to some practical observations.

PR #124, opened 2024-04-02
issue: 3844385 Fix new TCP timers registration lock contention

AlexanderGrissik:
## Description
Avoid calling register_socket_timer_event when a socket is already registered (TIME-WAIT). Although there is no functionality issue with that, it produces too high rate of posting events for internal-thread. This leads to lock contention inside internal-thread and degraded performance of HTTP CPS.

##### Why ?
HTTP CPS degradation.

## Change type
- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list

### DIFF
--- a/src/core/sock/sockinfo_tcp.cpp
+++ b/src/core/sock/sockinfo_tcp.cpp
@@ -2482,10 +2482,15 @@ ssize_t sockinfo_tcp::rx(const rx_call_t call_type, iovec *p_iov, ssize_t sz_iov
 
 void sockinfo_tcp::register_timer()
 {
-    si_tcp_logdbg("Registering TCP socket timer: socket: %p, thread-col: %p, global-col: %p", this,
-                  get_tcp_timer_collection(), g_tcp_timers_collection);
+    // A reused time-wait socket wil try to add a timer although it is already registered.
+    // We should avoid calling register_socket_timer_event unnecessarily because it introduces
+    // internal-thread locks contention.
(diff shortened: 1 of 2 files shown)


src/core/sock/sockinfo_tcp.cpp
```diff
@@ -6017,15 +6022,16 @@ void tcp_timers_collection::add_new_timer(sockinfo_tcp *sock)
...
+    if (sock->is_timer_registered()) {
+        __log_warn("Trying to add timer twice for TCP socket %p", sock);
         return;
     }
 
     sock_list &bucket = m_p_intervals[m_n_next_insert_bucket];
     bucket.emplace_back(sock);
     m_sock_remove_map.emplace(sock, std::make_tuple(m_n_next_insert_bucket, --(bucket.end())));
     m_n_next_insert_bucket = (m_n_next_insert_bucket + 1) % m_n_intervals_size;
+    sock->set_timer_registered(true);
```
iftahl, line 6034: Consider set it to true before putting in the bucket.
Not mandatory.

iftahl (approved):
Approved, just consider comment.

src/core/sock/sockinfo_tcp.cpp
AlexanderGrissik: As discussed, we leave it here as this should not affect functionality.

PR #145, opened 2024-05-05
issue: 3795997 Allow split segment with unacked q

iftahl:
When the send window is not big enough for the required TCP segment to send, we may split the segment so it will fit into the window. Before this change - We didn't split the segment in the case we have unacked segments. The motivation was that we anticipate to get ACK on the inflight segments, which will trigger the next send operation.
This flow counts on RTT for receiving ACKs, which may be delayed depending on the remote side. When RTT is long - we would block sending although the TCP send window allows it.

The change is to split TCP segments although we have unacked data, in case the send window is big enough (mss).

## Change type
- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list


iftahl:
bot:retest

iftahl:
@AlexanderGrissik can we merge it?

AlexanderGrissik (approved):

PR #180, opened 2024-07-09
issue: 3594311 Fix segfault for delegate incoming handshake failure 

AlexanderGrissik:
## Description
Seg fault fix for delegated timers and cleanups

##### What
See commit comments. 

##### Why ?
Crash fix for Nginx Proxy

## Change type
- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list

### DIFF
--- a/src/core/sock/sockinfo_tcp.cpp
+++ b/src/core/sock/sockinfo_tcp.cpp
@@ -292,7 +292,6 @@ sockinfo_tcp::sockinfo_tcp(int fd, int domain)
     , m_tcp_con_lock(get_new_tcp_lock())
     , m_sysvar_buffer_batching_mode(safe_mce_sys().buffer_batching_mode)
     , m_sysvar_tx_segs_batch_tcp(safe_mce_sys().tx_segs_batch_tcp)
-    , m_sysvar_tcp_ctl_thread(safe_mce_sys().tcp_ctl_thread)
     , m_tcp_seg_list(nullptr)
     , m_sysvar_rx_poll_on_tx_tcp(safe_mce_sys().rx_poll_on_tx_tcp)
     , m_user_huge_page_mask(~((uint64_t)safe_mce_sys().user_huge_page_size - 1))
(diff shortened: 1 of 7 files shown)


src/core/sock/sockinfo_tcp.cpp
```diff
@@ -1850,14 +1840,16 @@ void sockinfo_tcp::abort_connection()
     tcp_abort(&(m_pcb));
 }
 
-int sockinfo_tcp::handle_child_FIN(sockinfo_tcp *child_conn)
+void sockinfo_tcp::handle_incoming_handshake_failure(sockinfo_tcp *child_conn)
 {
-    lock_tcp_con();
+    child_conn->unlock_tcp_con();
```
iftahl, line 1843: why don't we leave the child_conn locked through the whole function?

iftahl (commented):
and regarding:
The handling of closing SYN_RCVD socket as part of handling FIN in sockinfo_tcp was removed, since we ignore FIN packets in SYN_RCVD state.

This socket closure will be done as part of "tcp_slowtmr - SYN_RCVD timeout"?

AlexanderGrissik:
bot:retest

src/core/sock/sockinfo_tcp.cpp
AlexanderGrissik: Reminder: Remove accpted_conn loop.

AlexanderGrissik: Removed unneeded m_accepted_conn loop after setting m_parent=nullptr before putting the conenction into m_accepted_conn

AlexanderGrissik: We do need to unlock child because we lock listen socket. It will work without, but it is not a good practice to have two locks acquired by the same thread.

AlexanderGrissik:
The case of FIN at SYN_RCVD state will be handled on timer. This is our current behavior until we decide to change that. RFC/Web do not state what exact behavior should implementation take in this case.

AlexanderGrissik:
bot:retest

iftahl (approved):

PR #215, opened 2024-08-14
issue: 4023262 Handle down interface states during table update

BasharRadya:
In this commit, we address a scenario where more than one network interface is assigned the same IP address. The current implementation supports creating RX rules for only one interface per IP address, which can lead to issues if multiple interfaces share the same IP.

Key Changes:

Added a mechanism to detect when the same IP address is assigned to multiple interfaces.
If this conflict occurs, a warning message is printed to notify the user that this configuration is not supported.
When faced with this conflict, the implementation now prioritizes the interface that is in the "Up" state, assigning the RX rules to it.
This change helps prevent potential conflicts and ensures that the most relevant (active) interface is used when an IP address conflict occurs.

## Change type
- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list

### DIFF
--- a/src/core/dev/net_device_table_mgr.cpp
+++ b/src/core/dev/net_device_table_mgr.cpp
@@ -190,6 +190,9 @@ void net_device_table_mgr::update_tbl()
     static int _seq = 0;
     net_device_val *p_net_device_val;
 
+    /* Track ips assigned to multiple-interfaces */
+    std::set<std::string> duplicate_ips;
+
     /* Set up the netlink socket */
     fd = SYSCALL(socket, AF_NETLINK, SOCK_RAW, NETLINK_ROUTE);
     if (fd < 0) {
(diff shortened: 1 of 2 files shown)


BasharRadya:
bot:retest

src/core/dev/net_device_table_mgr.cpp
```diff
@@ -264,19 +264,32 @@ void net_device_table_mgr::update_tbl()
...
+                    auto it = net_device_map.find(ip->local_addr);
+                    if (it == net_device_map.end() ||
+                        p_net_device_val->get_state() != net_device_val::DOWN) {
+                        net_device_map[ip->local_addr] = p_net_device_val;
+                    }
+                    if (it != net_device_map.end()) {
+                        vlog_printf(VLOG_WARNING,
+                                    "IP ADDRESS %s ASSIGNED TO MULTIPLE INTERFACES NOT SUPPORTED\n",
+                                    ip->local_addr.to_str(family).c_str());
+                    }
```
pasis, line 279: This approach can print multiple extra messages (per problematic interface). This is OK since this is unlikely scenario. We still can consider choosing among the following options:
1. Print a message per problematic interface as is. I'd prefer to move the if block at the top of the lambda not to think about iterator invalidation (the previous block manipulates with the collection).
2. Remember the fact that there is at least one duplicate IP and print a single common warning message at the end.
3. Create a map on stack with key=IP value=net_device_val and print a single message per IP at the end with mentioning which interface will be used for each IP.

Also rewrite the warning message:
 * Don't use upper case
 * Write more "light" message, because XLIO will work as expected in multiple scenarios even with duplicate IP addresses, and the current message makes impression that XLIO won't work at all. Use something like "Duplicate IP address %s is detected on multiple interfaces. XLIO will work with a single interface".

pasis (approved):

BasharRadya:
bot:retest

BasharRadya:
bot:retest

PR #251, opened 2024-10-13
Fix calculating TCP payload

BasharRadya:
Adjusted TCP payload length calculation by using the correct header size
to include TCP options.
The iovec is overridden later in the LWIP callback, but this ensures
correctness for future use.

## Change type
- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list


BasharRadya:
bot:retest

pasis:
This needs a close look, because TCP options can be handled as TCP payload in the code (for example, non-zc TX scenario keeps options in the 1st pbuf as part of payload)

PR #253, opened 2024-11-04
issue: 3914075 CQ Moderation with small payloads

tomerdbz:
Adapting interrupt rate per second to packet rate when using adaptive CQ moderation.

## Description
With Nginx scenarios when downloading small files (0B payloads) alongside XLIO_CQ_MODERATION_ENABLE=1 - a throughput degradation was observed.
This, in contrast to running the same scenarios with XLIO_CQ_MODERATION_ENABLE=0.

The degradation was 25% for 2400 connections, and 40% for 1200 connections.

Upon inspecting the performance, it seemed the adaptive Completion Event Moderation's `cq_timeout` and `cq_count` were extremely sensitive to momentarily spikes in CQEs - leading to overall degradation in the performance for the above use case.

##### What
Adapting interrupt rate to packet rate - be it short bursts or temporary silence, or other pattern changes - when using adaptive CQ moderation. This, in turn, changes the `cq_timeout` and `cq_count` values chosen for moderation.

##### Why ?
This PR solves bug 3914075.

##### How ?
I printed the arguments to `ring_simple::modify_cq_moderation`, witnessing them momentarily spiking.
upon ignoring the spikes by putting constant values, the performance improved.

We've witnessed the arguments are controlled by `XLIO_CQ_AIM_INTERRUPTS_RATE_PER_SEC` - which is 1000 by default.
I've conducted an extensive matrix of benchmarks - checking:
```
(1200 connections, 2400 connections) X (0B file, 100KB file) X (0 interrupt rate, 500 interrupt rate, 1000 interrupt rate, ..., 15000 interrupt rate)
```

The results were mixed, but it was clear every set of (connections, file_length) had a range of good interrupt rate values - and a set of really bad values.

The outcome was finding it dynamically, to adjust the interrupt rate based on the packet rate.
This gave best results across all use cases, i.e. - a second matrix of tests was conducted to test the proposed PR:
```
(1200 connections, 2400 connections) X (0B file, 100KB file) X (no_cq_moderation, cq_moderation_w/o_pr, cq_moderation_after_pr)
```

## Change type
- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

### DIFF
--- a/src/core/dev/ring_simple.cpp
+++ b/src/core/dev/ring_simple.cpp
@@ -122,6 +122,7 @@ ring_simple::ring_simple(int if_index, ring *parent, ring_type_t type, bool use_
     m_mtu = p_ndev->get_mtu();
 
     memset(&m_cq_moderation_info, 0, sizeof(m_cq_moderation_info));
+    m_cq_moderation_info.interrupt_rate_per_sec = safe_mce_sys().cq_aim_interrupts_rate_per_sec;
     memset(&m_tso, 0, sizeof(m_tso));
 #ifdef DEFINED_UTLS
     memset(&m_tls, 0, sizeof(m_tls));
@@ -1060,6 +1061,10 @@ void ring_simple::adapt_cq_moderation()
     BULLSEYE_EXCLUDE_BLOCK_END
 
     if (interval_packets == 0) {
+        // no traffic - wait eagerly for packet
+        // i.e. - bigger interrupt_rate_per_sec
+        m_cq_moderation_info.interrupt_rate_per_sec += 500;
+
         // todo if no traffic, set moderation to default?
         modify_cq_moderation(safe_mce_sys().cq_moderation_period_usec,
                              safe_mce_sys().cq_moderation_count);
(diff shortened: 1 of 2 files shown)


src/core/dev/ring_simple.cpp
```diff
@@ -1070,12 +1075,18 @@ void ring_simple::adapt_cq_moderation()
     uint32_t avg_packet_rate =
         (interval_packets * 1000) / (safe_mce_sys().cq_aim_interval_msec * (1 + missed_rounds));
 
-    uint32_t ir_rate = safe_mce_sys().cq_aim_interrupts_rate_per_sec;
+    uint32_t count = avg_packet_rate / m_cq_moderation_info.interrupt_rate_per_sec;
+    if (count > safe_mce_sys().cq_aim_max_count) {
+        // more traffic - wait patiently for packets and aggregate completions
+        // i.e. - smaller interrupt_rate_per_sec
+        m_cq_moderation_info.interrupt_rate_per_sec -= 500;
```
tomerdbz, line 1082: maybe return to default?

tomerdbz:
The PR was closed as it is out-of-scope for the bug fix.

TBD if we should adapt interrupt rate to packet pattern to get improvement - and how exactly to do so.

PR #265, opened 2024-11-18
issue: 4159515 Removing DOCA code from cq_mgr_tx

AlexanderGrissik:
## Description
1. Removed DOCA code from CQ mgr TX - CQ mgrs are planned to be removed
2. Removed mem_buf_tx_get inner polling   
3. Removed debug only variable in data path

##### What
Inner polling inside mem_buf_tx_get tries to mitigate bad app behavior or misconfiguration. 
XLIO focuses on HPC apps and it is better to avoid such unnecessary logics.
Additionally, the implementation introduces a race in working with notification mechanism.

##### Why ?
DOCA integration

## Change type
- [ ] Bugfix
- [X] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list

### DIFF
--- a/src/core/dev/ring_simple.cpp
+++ b/src/core/dev/ring_simple.cpp
@@ -542,127 +542,56 @@ int ring_simple::drain_and_proccess()
     return ret;
 }
 
-mem_buf_desc_t *ring_simple::mem_buf_tx_get(ring_user_id_t id, bool b_block, pbuf_type type,
-                                            int n_num_mem_bufs /* default = 1 */)
+mem_buf_desc_t *ring_simple::mem_buf_tx_get(ring_user_id_t id, pbuf_type type,
+                                            uint32_t n_num_mem_bufs /* default = 1 */)
 {
+    std::lock_guard<decltype(m_lock_ring_tx)> lock(m_lock_ring_tx);
     NOT_IN_USE(id);
-    int ret = 0;
-    mem_buf_desc_t *buff_list = nullptr;
-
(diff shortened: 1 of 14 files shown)


src/core/dev/ring_simple.cpp
```diff
@@ -750,11 +748,7 @@ bool ring_simple::is_available_qp_wr(bool b_block, unsigned credits)
                 if (p_cq_mgr_tx) {
 
                     // Allow additional CQ arming now
-                    if (safe_mce_sys().doca_tx) {
-                        m_hqtx->clear_notification();
-                    } else {
-                        p_cq_mgr_tx->reset_notification_armed();
-                    }
+                    p_cq_mgr_tx->reset_notification_armed();
```
iftahl, line 751: see in line 717:
"
            // TODO Resolve race window between previous polling and request_notification
            ret = m_p_cq_mgr_tx->request_notification();
"
also need to handle I think

AlexanderGrissik: I remove all the request_notification with polling mechanism for DOCA. These are just leftovers for dpcp path that will be removed with DPCP. I don't see a sense in all this. If there are no wqes then the sq is too small or it just always full in which case 99% this attempt will always fail. The same applies to the mem_buf_tx_get.
That's why i don't see a reason to fix the old path.

iftahl (approved):

AlexanderGrissik:
bot:retest

PR #270, opened 2024-12-02
issue: 4158085 DOCA TX credits

tomerdbz:
## Description
Using DOCA terminology to implement the equivalent to SW queue limits.

##### What
Limit the amount of TX task pool expansions.

##### Why ?
Solve https://redmine.mellanox.com/issues/4158085.

##### How ?

Utilizing the existing DOCA mechanisms for task pool limits and expansions, the following designs were evaluated:
1. A limit based on WQEBBs.
2. A limit based on in-flight bytes.

Both approaches required code duplication with doca_ether. Leveraging the existing DOCA task pool limitation mechanism achieves similar results without reinventing the wheel, albeit with reduced granularity and control.

## Change type
- [ ] Bugfix
- [X] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list

### DIFF
--- a/README
+++ b/README
@@ -88,6 +88,7 @@ Example:
  XLIO DETAILS: TCP max syn rate               0 (no limit)               [XLIO_TCP_MAX_SYN_RATE]
  XLIO DETAILS: Zerocopy Mem Bufs              200000                     [XLIO_ZC_BUFS]
  XLIO DETAILS: Zerocopy Cache Threshold       10 GB                      [XLIO_ZC_CACHE_THRESHOLD]
+ XLIO DETAILS: Tx Queue Max Elements          32K                        [XLIO_TX_QUEUE_MAX_ELEMENTS]
  XLIO DETAILS: Tx Mem Buf size                0                          [XLIO_TX_BUF_SIZE]
  XLIO DETAILS: ZC TX size                     32 KB                      [XLIO_ZC_TX_SIZE]
  XLIO DETAILS: Tx QP WRE                      32768                      [XLIO_TX_WRE]
@@ -313,6 +314,12 @@ Default value is 200000
 XLIO_ZC_CACHE_THRESHOLD
 Memory limit for the mapping cache which is use by sendfile().
 Default value is 10GB
+
+Name: XLIO_TX_QUEUE_MAX_ELEMENTS
+Defines the max queue size for TX queues.
+Higher values allow buffering more tasks internally, which improves NIC utilization. However, this may introduce delays in NIC recovery if a burst persists for extended periods.
+Lower values help the NIC to recover more quickly from bursts but may limit its capacity to efficiently manage large bursts of data.
+Default value is 32K
 
 XLIO_TX_BUF_SIZE
 Size of Tx data buffer elements allocation.
--- a/src/core/dev/hw_queue_tx.cpp
+++ b/src/core/dev/hw_queue_tx.cpp
@@ -334,6 +334,8 @@ bool hw_queue_tx::prepare_doca_txq()
         return false;
     }
 
+    m_task_list_count = max_burst_size;
+
     err = doca_eth_txq_set_l3_chksum_offload(txq, DOCA_CHECKSUM_HW_L3_ENABLE);
     if (DOCA_IS_ERROR(err)) {
         PRINT_DOCA_ERR(hwqtx_logerr, err, "doca_eth_txq_set_l3_chksum_offload txq: %p",
@@ -2001,18 +2003,30 @@ bool hw_queue_tx::expand_doca_inventory()
 
 bool hw_queue_tx::expand_doca_task_pool(bool is_lso)
 {
+    if (m_task_list_count >= safe_mce_sys().tx_queue_max_elements) {
+        hwqtx_logfunc("Silent packet drop, can't expand task pool");
+        return false;
+    }
+
+    const uint16_t expand_batch_size =
+        (m_task_list_count + DOCA_EXPAND_BATCH_SIZE) >= safe_mce_sys().tx_queue_max_elements
+        ? (m_task_list_count + DOCA_EXPAND_BATCH_SIZE) % DOCA_EXPAND_BATCH_SIZE
+        : (DOCA_EXPAND_BATCH_SIZE);
+
     doca_error_t rc;
     doca_eth_txq *txq = m_doca_txq.get();
     if (is_lso) {
-        rc = doca_eth_txq_task_lso_send_num_expand(txq, DOCA_EXPAND_BATCH_SIZE);
+        rc = doca_eth_txq_task_lso_send_num_expand(txq, expand_batch_size);
     } else {
-        rc = doca_eth_txq_task_send_num_expand(txq, DOCA_EXPAND_BATCH_SIZE);
+        rc = doca_eth_txq_task_send_num_expand(txq, expand_batch_size);
     }
 
     if (DOCA_IS_ERROR(rc)) {
         PRINT_DOCA_ERR(hwqtx_logerr, rc, "DOCA expand task pool (LSO=%d) failed", is_lso);
         return false;
     }
+
+    m_task_list_count += expand_batch_size;
     return true;
 }
 
--- a/src/core/util/sys_vars.cpp
+++ b/src/core/util/sys_vars.cpp
@@ -31,6 +31,7 @@
  * SOFTWARE.
  */
 
+#include "doca_log.h"
 #ifdef HAVE_CONFIG_H
 #include "config.h"
 #endif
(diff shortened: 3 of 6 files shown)


tomerdbz:
ready for review
@galnoam  @AlexanderGrissik

README
```diff
@@ -88,6 +88,7 @@ Example:
  XLIO DETAILS: TCP max syn rate               0 (no limit)               [XLIO_TCP_MAX_SYN_RATE]
  XLIO DETAILS: Zerocopy Mem Bufs              200000                     [XLIO_ZC_BUFS]
  XLIO DETAILS: Zerocopy Cache Threshold       10 GB                      [XLIO_ZC_CACHE_THRESHOLD]
+ XLIO DETAILS: Tx Allowed Task Pool Expands   0                          [XLIO_TX_ALLOWED_TASK_POOL_EXPANDS]
```
galnoam, line 91: Not sure customer will understand the functionality from the name.
I think its implementation detail "task pool expansion",
in my opinion it should be in pkt size or memory size, and we should convert it internally.
what do you think?

README
```diff
@@ -314,6 +315,14 @@ XLIO_ZC_CACHE_THRESHOLD
 Memory limit for the mapping cache which is use by sendfile().
 Default value is 10GB
 
+XLIO_TX_ALLOWED_TASK_POOL_EXPANDS
```
galnoam, line 318: same here if you agree with above

galnoam (changes requested):
@tomerdbz see if you agree with my comment

README
tomerdbz: You're super right. changed it - wdyt?

tomerdbz:
bot:retest

README
```diff
@@ -88,6 +88,7 @@ Example:
  XLIO DETAILS: TCP max syn rate               0 (no limit)               [XLIO_TCP_MAX_SYN_RATE]
  XLIO DETAILS: Zerocopy Mem Bufs              200000                     [XLIO_ZC_BUFS]
  XLIO DETAILS: Zerocopy Cache Threshold       10 GB                      [XLIO_ZC_CACHE_THRESHOLD]
+ XLIO DETAILS: Tx Burst Handling Limit        0                          [XLIO_TX_BURST_HANDLING_LIMIT]
```
galnoam, line 91: XLIO_TX_SW_QUEUE_EXPAND?
@tomerdbz WDYT?

iftahl (commented):
1. I would not recommend 0 as a default - it will probably have a performance impact. Did you check it? Another point is that using 0 will prevent us to utilize DOCA SW queue due to a low number of tasks we allocate initially.
Use a default that on the one hand will be big enough to utilize SQ queue, but on the other hand won't be too big, for example 64K tasks.
2. I would count the number of tasks, and not the number of expansions.

iftahl (changes requested):
See prev comments.

tomerdbz:
I'll check again. The current behavior in XLIO is no SW queue by default,
so the thought was to be in line with current behavior.
The initial number of tasks is a function of a API call to DOCA that
essentially correlates to the HW queue length.

As for counting tasks instead of expansions. Why is it better?

galnoam:
@tomerdbz, I don't think that the user will understand what tasks is as well.
Let's go with XLIO_TX_SW_QUEUE_EXPAND.

@iftahl:
1. We decided to keep the current XLIO behavior, no SW queue. is that correct? that we maintain the current state this way?
2. Agree that it's better to give more precise configuration, if possible, we should do it but should on be in terms of tasks. but ether size in bytes or pkts. user do not know what tasks is.

iftahl:
First, user will not understand expands just like tasks...
Second, with 0 expands we will have performance issues because we will not be able to utilize all sq. This is why we demanded new api for expansions...
Third, there is no direct relation to sw queue at all so the name of the parameter is not correct... You will never know if you used the hw or the sw queue of doca... Its internal.

I would just make a param that controls the max number of tasks, with a good description, and perfomance tests for this pr is a must.




Thanks,

Iftah Levi
Senior SW Engineer | NVIDIA<http://www.nvidia.com/>

________________________________

galnoam:
@iftahl
- not agree about expand, it may be not understood how to use it, but what it does will be understood.
- the idea is to maintain current XLIO functionality, if it's not then it a problem, so I agree we should avoid degradation.
  how can we get the HW q size? there is no way to do it?
- Moshe mentioned few times that we should avoid using their SW queue, performance wise.
  Are you sure we do not have a way to get HW q size?
- I don't want to use task in the name, its implementation detail.

iftahl:
PSB


Thanks,

Iftah Levi
Senior SW Engineer | NVIDIA<http://www.nvidia.com/>



________________________________

galnoam:
@Tomer ***@***.***>, lets discuss @Iftah ***@***.***> comments.

Thanks,
Gal

________________________________

src/core/util/sys_vars.cpp
```diff
@@ -1149,6 +1153,10 @@ void mce_sys_var::get_env_params()
         zc_cache_threshold = option_size::from_str(env_ptr);
     }
 
+    if ((env_ptr = getenv(SYS_VAR_TX_QUEUE_EXPANSIONS))) {
```
iftahl, line 1156: make the env similar to our internal sys var, it is weird that one contains burst for example, and the other one does not...

src/core/dev/hw_queue_tx.cpp
```diff
@@ -2013,6 +2018,8 @@ bool hw_queue_tx::expand_doca_task_pool(bool is_lso)
         PRINT_DOCA_ERR(hwqtx_logerr, rc, "DOCA expand task pool (LSO=%d) failed", is_lso);
         return false;
     }
+
+    ++m_task_list_expand_count;
```
iftahl, line 2022: 1. Each expansion expands by "DOCA_EXPAND_BATCH_SIZE", not 1.
2. Please use 32K convention and not 32000, see other params for reference.

iftahl (changes requested):

src/core/dev/hw_queue_tx.cpp
tomerdbz: Thanks @iftahl. very much appreciated :)
Fixed comments. let me know if it's OK by you.

src/core/dev/hw_queue_tx.cpp
```diff
@@ -2013,6 +2018,8 @@ bool hw_queue_tx::expand_doca_task_pool(bool is_lso)
         PRINT_DOCA_ERR(hwqtx_logerr, rc, "DOCA expand task pool (LSO=%d) failed", is_lso);
         return false;
     }
+
+    m_task_list_expand_count += DOCA_EXPAND_BATCH_SIZE;
```
AlexanderGrissik, line 2022: See my comment about naming below. In this case, we better call it something like m_task_list_size

AlexanderGrissik: maybe we can even fetch that number from doca to avoid keeping a variable to track it.

tomerdbz: Thanks. looked and no API unfortunately. as per your name comment - fixed.

src/core/util/sys_vars.h
```diff
@@ -719,6 +721,7 @@ extern mce_sys_var &safe_mce_sys();
 #define MCE_DEFAULT_TCP_MAX_SYN_RATE         (0)
 #define MCE_DEFAULT_TCP_NODELAY_TRESHOLD     (0)
 #define MCE_DEFAULT_ZC_CACHE_THRESHOLD       (10LU * 1024 * 1024 * 1024) // 10GB
+#define MCE_DEFAULT_TX_QUEUE_EXPANSIONS      (32)
```
AlexanderGrissik, line 724: this will break if we change the hardcoded DOCA_EXPAND_BATCH_SIZE

tomerdbz:
I really don't mind folks as this is just step 0 in a bigger process.

But the expansion thing was decided with Gal after talking to our Technical
writer.

We chose expansions over total size. We can change it but please let's make
sure this time we are all aligned.

galnoam:
XLIO_TX_QUEUE_MAX_ELEMENTS 
the defaults should be as discussed.
the user will set the exact num of elements he needs and not num of expansions.
we can still do lazy init if needed and start with lower number and use expansions till the limit.

tomerdbz:
@AlexanderGrissik @iftahl @galnoam  Done :)

galnoam:
@iftahl, @AlexanderGrissik  please check and approve if comments fixed.

src/core/dev/hw_queue_tx.cpp
```diff
@@ -2001,18 +2003,30 @@ bool hw_queue_tx::expand_doca_inventory()
 
 bool hw_queue_tx::expand_doca_task_pool(bool is_lso)
 {
+    if (m_task_list_count >= safe_mce_sys().tx_queue_max_elements) {
+        hwqtx_logfunc("silent packet drop, can't expand task pool!");
```
AlexanderGrissik, line 2007: Please use capital letters in the start of logging even they are logfunc. Also no need for marks such as '!'.

src/core/util/sys_vars.h
```diff
@@ -719,6 +721,7 @@ extern mce_sys_var &safe_mce_sys();
 #define MCE_DEFAULT_TCP_MAX_SYN_RATE         (0)
 #define MCE_DEFAULT_TCP_NODELAY_TRESHOLD     (0)
 #define MCE_DEFAULT_ZC_CACHE_THRESHOLD       (10LU * 1024 * 1024 * 1024) // 10GB
+#define MCE_DEFAULT_TX_QUEUE_MAX_ELEMENTS    (32000)
```
AlexanderGrissik, line 724: Why not 32768?

AlexanderGrissik (approved):

PR #291, opened 2025-01-05
issue: 4213201 Fix locking on TX path

tomerdbz:
## Description

In high-throughput environments like PCIe Gen5, a race condition was causing a double-free error
in the tcp_tx_pbuf_free function.

The lwip_pbuf reference count was being decremented without proper locking mechanisms in place.
This lack of synchronization could lead to multiple threads decrementing the reference count simultaneously,
causing it to reach zero prematurely.

##### What
Introduced a recursive mutex pbuf_ref_count_lock to ensure that the decrement and check of the lwip_pbuf reference count are thread-safe.

The lock is declared as static to ensure it is shared across all instances and prevent multiple threads from entering the critical section simultaneously.

##### Why ?
Fix bug https://redmine.mellanox.com/issues/4213201

##### How ?
1. sniffed and seen RSTs. seen in the error log that just before getting the double free message we get a message that the "client closed connection while waiting for request".
2. Tried to debug, yet while debugging the race hadn't re-occurred.
3. added a print_stacktrace logic on the double-free to see the entire context (that wasn't related to the RST).
4. added a print_stacktrace logic call to tcp_rst to see if it's called before the double-free - it was indeed the case.
5. inspected the stack-trace for proper locking - challenging the assumptions of the allocator logic.
6. Found it :)

## Change type
- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list

### DIFF
--- a/src/core/sock/sockinfo_tcp.cpp
+++ b/src/core/sock/sockinfo_tcp.cpp
@@ -5692,18 +5692,24 @@ void sockinfo_tcp::tcp_tx_pbuf_free(void *p_conn, struct pbuf *p_buff)
     sockinfo_tcp *p_si_tcp = (sockinfo_tcp *)(((struct tcp_pcb *)p_conn)->my_container);
     dst_entry_tcp *p_dst = (dst_entry_tcp *)(p_si_tcp->m_p_connected_dst_entry);
 
+    static lock_mutex_recursive pbuf_ref_count_lock("sockinfo_tcp::tcp_tx_pbuf_free");
     if (likely(p_dst)) {
         p_dst->put_buffer((mem_buf_desc_t *)p_buff);
     } else if (p_buff) {
         mem_buf_desc_t *p_desc = (mem_buf_desc_t *)p_buff;
 
-        // potential race, ref is protected here by tcp lock, and in ring by ring_tx lock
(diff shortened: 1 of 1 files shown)


src/core/sock/sockinfo_tcp.cpp
```diff
@@ -5692,18 +5692,24 @@ void sockinfo_tcp::tcp_tx_pbuf_free(void *p_conn, struct pbuf *p_buff)
     sockinfo_tcp *p_si_tcp = (sockinfo_tcp *)(((struct tcp_pcb *)p_conn)->my_container);
     dst_entry_tcp *p_dst = (dst_entry_tcp *)(p_si_tcp->m_p_connected_dst_entry);
```
tomerdbz, line 5694: I thought of 2 alternatives here:
1. a global static - with eager initialization - doable.
2. a member in the class - enlarging the already large instance - didn't like it - but doable.
3. lazy initialization with a local static - felt more readable (yet debateable) - yet you do pay a small perf price on first call  to tcp_tx_pbuf_free

tomerdbz:
Draft as setup was "hijacked" late w/o verification ask. will ask and make sure this is the correct setup.

PR #307, opened 2025-02-06
issue: 4287937 Fix a crash on shrinking a FIN segment with data

pasis:
## Description
This fixes 2 bugs that lead to a crash if combined:
1. ip_output() converts a pbuf object from stack to mem_buf_desc_t which leads to reading of garbage;
2. Acknowledging all the data in a FIN segment but not the FIN flag itself leads to removing all the pbufs from the segments and NULL pointer dereference at the end.

##### What
Fix a crash.

##### Why ?
Bugfixes.

## Change type
- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list

### DIFF
--- a/src/core/lwip/tcp_in.c
+++ b/src/core/lwip/tcp_in.c
@@ -947,6 +947,15 @@ static void ack_partial_or_whole_segment(struct tcp_pcb *pcb, u32_t ackno, struc
     struct tcp_seg *whole_seg_to_ack;
     while ((*seg) != NULL && TCP_SEQ_GT(ackno, (*seg)->seqno)) {
         if (TCP_SEQ_LT(ackno, (*seg)->seqno + TCP_SEGLEN((*seg)))) {
+            if ((*seg)->tcp_flags & TCP_FIN) {
+                // Avoid shrinking a segment with the FIN flag not to handle corner cases.
+                // The most challenging scenario is ACK for the entire data, but not the FIN. In
+                // this case the shrink implementation removes all the pbufs from the segment
+                // what leads to a segfault eventually.
+                // Let's keep the whole segment and retransmit the duplicate data if needed
+                // until the shrink implementations are improved or removed by arch update.
+                break;
+            }
             // Ack partial TCP segment
             u32_t removed = (*seg)->flags & TF_SEG_OPTS_ZEROCOPY
                 ? tcp_shrink_zc_segment(pcb, (*seg), ackno)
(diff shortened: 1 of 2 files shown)


src/core/lwip/tcp_in.c
```diff
@@ -947,6 +947,15 @@ static void ack_partial_or_whole_segment(struct tcp_pcb *pcb, u32_t ackno, struc
     struct tcp_seg *whole_seg_to_ack;
     while ((*seg) != NULL && TCP_SEQ_GT(ackno, (*seg)->seqno)) {
         if (TCP_SEQ_LT(ackno, (*seg)->seqno + TCP_SEGLEN((*seg)))) {
+            if ((*seg)->tcp_flags & TCP_FIN) {
```
tomerdbz, line 950: 1. the fact we retransmit all the data - even if it was partially ACKED is OK and the comment is good but imho should be documented in redmine as well as a known issue.

Suggestion for the issue I'd add:
Splitting the last FIN-bearing segment into two (data-only and FIN-only segments) can allow partial ACK to safely remove the data portion while preserving the FIN portion. requires a deeper refactor of how segments are created and queued, but will make the session more efficient.

2. if there's a window update (e.g. - due to congestion) for some reason, can retransmitting all the data be a problem that will stall us?

src/core/sock/sockinfo_tcp.cpp
```diff
@@ -1393,7 +1393,7 @@ err_t sockinfo_tcp::ip_output(struct pbuf *p, struct tcp_seg *seg, void *v_p_con
     void *cur_end;
 
     if (unlikely(flags & XLIO_TX_PACKET_REXMIT)) {
-        if (unlikely(inspect_socket_error_state(reinterpret_cast<const mem_buf_desc_t *>(p),
+        if (unlikely(inspect_socket_error_state(reinterpret_cast<const mem_buf_desc_t *>(seg->p),
```
tomerdbz, line 1396: I think we should not ignore the investigation of why this bug caused the shrink w/o FIN bug. 
it's not obvious to me - and I think the effort will either uncover more bugs, or lead to gaining a very important insight.

tomerdbz (changes requested):
A patch with a good tradeoff of (time and correctness) VS. (optimization).
Great work uncovering that! :)

tomerdbz (approved):

PR #317, opened 2025-03-05
align xlio_socket_api.c with updated implementation

changchengx:
## Description
Update ```tests/extra_api/xlio_socket_api.c``` implementation. Extra APIs are called with using ```xlio_get_api```

##### What
_Subject: ```align xlio_socket_api.c with updated implementation```

##### Why ?
Without changing, the xlio_socket_api.c can't build/linked with libxlio.so

##### How ?
Call extra API with using xlio_get_api.

## Change type
- [ ] Bugfix
- [ ] Build related changes


galnoam:
@pasis, do we need this PR?

galnoam:
> @pasis, do we need this PR?

@pasis

pasis:
@galnoam This PR converts the manual test from statically linked to LD_PRELOAD version. They're just 2 options how we can integrate Ultra API into an application. The question is what option we choose.

P.S. XLIO doesn't support linking against shared library, so the official way is either linking against static lib or LD_PRELOAD of the shared lib

PR #326, opened 2025-03-31
issue: 4347777 Replace thread-local dummy lock with global one

tomerdbz:
## Description
This change replaces a thread-local dummy locker with a global one in ring_slave.cpp to fix a use-after-free issue.

#### What
Replace thread-local dummy locker with a global one in ring_slave.

#### Why ?
During XLIO shutdown, when one thread attempts to access a socket's locker that was created by a terminated thread, a use-after-free issue occurs because the thread-local dummy locker object is freed when its creator thread terminates. This can lead to segmentation faults during cleanup.

#### How ?
The solution is straightforward:
1. Replace static thread_local lock_dummy t_lock_dummy_ring with a global static lock_dummy g_lock_dummy_ring
2. Update get_new_lock() to use this global locker
4. The global locker is initialized with a name for better debugging: "global_dummy_ring"

This change ensures the dummy locker remains valid throughout the program's lifetime, regardless of thread termination status.

#### Performance considerations

Pros of global implementation:
1. The dummy lock is accessed frequently in the data path
2. Each access to a thread-local variable requires the more complex memory access pattern
3. This compounds the overhead across many accesses

The global was aligned to be in a different cache line to prevent false-sharing - thus not introducing perf hit,

#### Tests
added a system-test (gtest) that discovered same issue occured on src/core/sock/sockinfo_tcp.cpp.
Applied the same fix there as well for the test to pass.

## Change type
- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [X] Tests
- [ ] Other

## Check list

### DIFF
--- a/src/utils/lock_wrapper.h
+++ b/src/utils/lock_wrapper.h
@@ -473,6 +473,14 @@ public:
     int is_locked_by_me() override { return 1; }
 };
 
+// Users of lock_dummy may wish to alignas(64) to place this lock in in a different cache-line and
+// prevent false sharing
+struct alignas(64) padded_lock_dummy {
+    lock_dummy lock;
+    // Padding to fill a full cache line
+    char padding[64 - sizeof(lock_dummy)];
+};
+
 static inline void lock_deleter_func(lock_base *lock)
 {
     lock->delete_obj();
--- a/src/core/dev/ring_slave.cpp
+++ b/src/core/dev/ring_slave.cpp
@@ -21,13 +21,13 @@
 // AF_INET address 0.0.0.0:0, used for 3T flow spec keys.
 static const sock_addr s_sock_addrany;
 
-static thread_local lock_dummy t_lock_dummy_ring;
+static padded_lock_dummy g_lock_dummy_ring;
 
 static lock_base *get_new_lock(const char *name, bool real_lock)
 {
     return (real_lock
                 ? static_cast<lock_base *>(multilock::create_new_lock(MULTILOCK_RECURSIVE, name))
-                : static_cast<lock_base *>(&t_lock_dummy_ring));
+                : static_cast<lock_base *>(&g_lock_dummy_ring.lock));
 }
 
 ring_slave::ring_slave(int if_index, ring *parent, ring_type_t type, bool use_locks)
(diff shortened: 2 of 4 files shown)


tomerdbz:
this PR solves the issue - but it has some perf cost:
Thread-local storage typically has better cache locality since each thread has its own copy.

src/utils/lock_wrapper.h
```diff
@@ -1,35 +1,7 @@
...
- * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
- * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
- * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
- * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS
- * BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
- * ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
- * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
- * SOFTWARE.
+ * Copyright (c) 2021-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: GPL-2.0-only or BSD-2-Clause
```
BasharRadya, line 4: this is already updated in vnext branch - https://github.com/Mellanox/libxlio/blob/vNext/src/utils/lock_wrapper.h

you are working on top of old vnext - update it

AlexanderGrissik: In latest vNext we still have the thread_local
https://github.com/Mellanox/libxlio/blob/vNext/src/core/sock/sockinfo_tcp.cpp#L60

src/utils/lock_wrapper.h
```diff
@@ -550,4 +522,4 @@ class multilock {
     std::unique_ptr<lock_base, lock_deleter> m_lock;
 };
 
-#endif // LOCK_WRAPPER_H
+#endif // LOCK_WRAPPER_H
```
AlexanderGrissik, line 525: Missing end of line

AlexanderGrissik (approved):

BasharRadya (approved):

BasharRadya (approved):

src/utils/lock_wrapper.h
tomerdbz: fixed

src/utils/lock_wrapper.h
tomerdbz: fixed

tomerdbz:
bot:retest

galnoam:
@tomerdbz , please move to Fix/closed

PR #334, opened 2025-04-09
issue: 4398221 Fix connection closure on SYN_RCVD

tomerdbz:
## Description

When closing a connection in SYN_RCVD state, we should transition
directly to CLOSED without sending FIN or RST packets, as specified
in the RFC. This patch properly handles this case by immediately
changing the state to CLOSED instead of going through the normal
shutdown procedure.

This prevents unnecessary control packets when a connection receives
a SYN and is immediately closed without exchanging data.

##### What
according to the RFC, in case we get a SYN and no more data we should just close w/o FIN or RST.
this PR does that.

##### How ?
Found it as part of working on https://redmine.mellanox.com/issues/4050516

## Change type
- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list


tomerdbz:
to trigger the bug on any environment:

1. compile the following program
`#include <arpa/inet.h>
#include <netinet/in.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <sys/socket.h>
#include <sys/types.h>
#include <unistd.h>

int main() {
  int server_fd;
  struct sockaddr_in address;
  int opt = 1;
  int addrlen = sizeof(address);

  printf("TCP server cycling program starting...\n");

  // Create socket file descriptor
  if ((server_fd = socket(AF_INET, SOCK_STREAM, 0)) < 0) {
    perror("socket failed");
    exit(EXIT_FAILURE);
  }

  // Set socket options
  if (setsockopt(server_fd, SOL_SOCKET, SO_REUSEADDR | SO_REUSEPORT, &opt,
                 sizeof(opt))) {
    perror("setsockopt failed");
    close(server_fd);
    exit(EXIT_FAILURE);
  }

  // Setup server address structure
  memset(&address, 0, sizeof(address));
  address.sin_family = AF_INET;
  address.sin_addr.s_addr = INADDR_ANY;
  address.sin_port = htons(55400);

  // Bind socket to the port
  if (bind(server_fd, (struct sockaddr *)&address, sizeof(address)) < 0) {
    perror("bind failed");
    close(server_fd);
    exit(EXIT_FAILURE);
  }

  // Listen for connections
  if (listen(server_fd, 3) < 0) {
    perror("listen failed");
    close(server_fd);
    exit(EXIT_FAILURE);
  }

  while (1) {
    // Accept a connection
    int new_socket;
    struct sockaddr_in client_addr;
    socklen_t client_addrlen = sizeof(client_addr);
    if ((new_socket = accept(server_fd, (struct sockaddr *)&client_addr,
                             &client_addrlen)) < 0) {
      perror("accept failed");
      close(server_fd);
      exit(EXIT_FAILURE);
    }
    // Close the accepted connection
    close(new_socket);
  }

  // This point will never be reached in the current code
  close(server_fd);
  return 0;
}`

2. syn-flood it from a client
3. bug will trigger on vNext. after the patch you could see it's fixed :)

src/core/lwip/tcp.c
```diff
@@ -258,8 +258,16 @@ err_t tcp_close(struct tcp_pcb *pcb)
...
-    return tcp_close_shutdown(pcb, 1);
+
+    if (get_tcp_state(pcb) == SYN_RCVD) {
+        // according to the RFC, in case we get a SYN and no more data
+        // we should just close w/o FIN or RST
+        tcp_pcb_purge(pcb);
+        set_tcp_state(pcb, CLOSED);
+        return ERR_OK;
+    } else {
+        return tcp_close_shutdown(pcb, 1);
```
BasharRadya, line 269: inside tcp_close_shutdown we can see this code(below), we can fix the logic in "case SYN_RCVD" instead of adding this if condition.

    switch (get_tcp_state(pcb)) {
...
    case SYN_RCVD:
        err = tcp_send_fin(pcb);
        if (err == ERR_OK) {
            set_tcp_state(pcb, FIN_WAIT_1);
        }
..

tomerdbz: Fixed. thanks :) note that I needed to add an if before the switch-case as I don't want to address rst_on_unacked_data.

BasharRadya (approved):

src/core/lwip/tcp.c
```diff
@@ -148,6 +148,14 @@ static err_t tcp_close_shutdown(struct tcp_pcb *pcb, u8_t rst_on_unacked_data)
...
     err_t err;
 
+    if (get_tcp_state(pcb) == SYN_RCVD) {
+        // according to the RFC, in case we get a SYN and no more data
+        // we should just close w/o FIN or RST
+        tcp_pcb_purge(pcb);
+        set_tcp_state(pcb, CLOSED);
+        return ERR_OK;
+    }
+
```
pasis, line 158: The RFC states that FIN needs to be sent in this case. So, looks like original code makes sense.

[section 3.10.4](https://datatracker.ietf.org/doc/html/rfc9293#name-close-call)
```
3.10.4.  CLOSE Call
...
SYN-RECEIVED STATE

If no SENDs have been issued and there is no pending data to send, then form a FIN segment and send it, and enter FIN-WAIT-1 state; otherwise, queue for processing after entering ESTABLISHED state.
```

tomerdbz: You're right @pasis.
If we get a timeout we should just abandon.
If an app closes an accepted socket, we should initiate FIN flow.

2 issues:
1. Currently, on vNext - closing an accepted socket just sends a fin and closes the socket (calls internal thread with UNREGISTER_TCP_SOCKET_TIMER_AND_DELETE which destructs it). this doesn't conform too to the RFC!
2. there's no real distinguishing between app close to xlio initiating the close. See `handle_incoming_handshake_failure` - it actively calls the close syscall if we get a timeout - hence, we can't distinguish really between the two cases.

This makes the bug a much harder to fix - but I'm up to the challenge :) let me know if you got me and we're on the same page.

galnoam:
@pasis , please review.
If approved, is it relevant to VMA as well?

src/core/lwip/tcp.c
tomerdbz: @pasis 
Actually, I instrumented the code, adding this to tcp.c line 760:
printf("DEBUG: tcp_slowtmr PATH1 BEFORE purge: state=%d, unsent=%p unacked=%p\n",
                   get_tcp_state(pcb), pcb->unsent, pcb->unacked);

it validated we are aborting due to being "too long in SYN-RCVD" - we are not in CLOSE flow (3.10.4).

I've tried to look online what to do on timeout on SYN-RCVD, but it's not well-defined.
the linux kernel though simply abandons w/o sending RST, w/o sending FIN, as we do here - so I believe this solution is correct

tomerdbz: but this instrumentation does show the fix is not quite right - let me alter it

tomerdbz:
bot:retest

galnoam:
@tomerdbz , @BasharRadya , please resolve all comments

PR #337, opened 2025-04-22
issue: 4409403 Fix heap corruption since c73d96a

tomerdbz:
## Description
This PR fixes a critical heap corruption issue in TCP socket timer management
that was introduced in commit c73d96a3e7a9b64c77ed024fa94cf00b7fdca339.

##### What
Fix race condition between timer thread and socket deletion in TCP socket timers.

##### Why ?
A race condition was introduced in commit c73d96a that allowed the timer thread
to access socket objects after they had been deleted by the event handler thread,
causing heap corruption and crashes.

##### How ?
The fix improves synchronization between threads:
1. Removes sockets from timer collections while holding the socket lock
2. Creates a new deletion path that doesn't attempt to access timer collections
after socket cleanup
3. Additionally fixes a lock leak in the early return path of clean_socket_obj()

## Change type
- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list

## How this bug was found?
1. sockperf UDP ping-pong mode recreated this issue - it was found during new config STP.
2. git bisect found the commit
3. Cursor helped with explaining the commit issues
4. bug found

attaching a pic for a sequence diagram explaining the faulty flow that was fixed:
image

### DIFF
--- a/src/core/event/event_handler_manager.cpp
+++ b/src/core/event/event_handler_manager.cpp
@@ -156,12 +156,12 @@ void event_handler_manager::unregister_timers_event_and_delete(timer_handler *ha
     post_new_reg_action(reg_action);
 }
 
-void event_handler_manager::unregister_socket_timer_and_delete(sockinfo_tcp *sock_tcp)
+void event_handler_manager::unregister_socket_and_delete(sockinfo_tcp *sock_tcp)
 {
-    evh_logdbg("Unregistering TCP socket timer: %p", sock_tcp);
+    evh_logdbg("Deleting TCP socket: %p", sock_tcp);
     reg_action_t reg_action;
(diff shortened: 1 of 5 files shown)


pasis:
As we discussed offline,
The "timer thread" shouldn't exist and running timers outside the internal thread (event handler thread) is problematic indeed. In the past, we removed such a flow when the tcp timer could run as part of sockinfo_tcp::tcp_con_unlock() causing issues.

First, we need to identify the flow to run tcp timer outside of the internal thread context and this flow will like need to be avoided.

Note, this comment assumes that delegated timers feature is off.

PR #342, opened 2025-04-23
issue: 4446408 Add openssl version to the debug logs

pasis:
## Description
TLS RX offload depends on openssl library for SW encryption in case of a resync flow. Unless application is linked against openssl, XLIO cannot provide TLS RX offload support. Therefore, it's crucial to understand the environment XLIO works with if debug logs are provided offline.

##### What
Add openssl version to the debug logs.

##### Why ?
Improve debuggability.

## Change type
- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [x] Other

## Check list

### DIFF
--- a/src/core/sock/sockinfo_ulp.cpp
+++ b/src/core/sock/sockinfo_ulp.cpp
@@ -80,6 +80,8 @@ struct xlio_tls_api {
                               const unsigned char *);
     int (*EVP_EncryptUpdate)(EVP_CIPHER_CTX *, unsigned char *, int *, const unsigned char *, int);
     int (*EVP_EncryptFinal_ex)(EVP_CIPHER_CTX *, unsigned char *, int *);
+    const char *(*OpenSSL_version)(int t);
+    unsigned long (*OpenSSL_version_num)();
 };
 
 static struct xlio_tls_api *g_tls_api = nullptr;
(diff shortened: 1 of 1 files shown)


galnoam:
@pasis , lets create a ticket and add to commit tilte

src/core/sock/sockinfo_ulp.cpp
```diff
@@ -104,12 +106,20 @@ void xlio_tls_api_setup()
...
         s_tls_api.EVP_CIPHER_CTX_reset && s_tls_api.EVP_aes_128_gcm && s_tls_api.EVP_aes_256_gcm &&
         s_tls_api.EVP_DecryptInit_ex && s_tls_api.EVP_DecryptUpdate &&
         s_tls_api.EVP_CIPHER_CTX_ctrl && s_tls_api.EVP_DecryptFinal_ex &&
         s_tls_api.EVP_EncryptInit_ex && s_tls_api.EVP_EncryptUpdate &&
         s_tls_api.EVP_EncryptFinal_ex) {
+        if (s_tls_api.OpenSSL_version) {
+            vlog_printf(VLOG_DEBUG, "Found OpenSSL version: %s\n", s_tls_api.OpenSSL_version(0));
+        } else if (s_tls_api.OpenSSL_version_num) {
+            vlog_printf(VLOG_DEBUG, "Found OpenSSL version: 0x%lx",
+                        s_tls_api.OpenSSL_version_num());
```
BasharRadya, line 121: maybe print more readable output instead fo hexa if OpenSSL_version_num was used
unsigned long v=OpenSSL_version_num();
    int major = (v >> 28) & 0xF;
    int minor = (v >> 20) & 0xFF;
    int fix   = (v >> 12) & 0xFF;
    int patch = (v >> 4)  & 0xFF;
    int status= v & 0xF;

    printf("OpenSSL version: %d.%d.%d (patch %d) %s\n",
        major, minor, fix, patch,
        (status == 0) ? "release" : "pre-release/dev");

src/core/sock/sockinfo_ulp.cpp
```diff
@@ -104,12 +106,20 @@ void xlio_tls_api_setup()
...
+    XLIO_TLS_API_FIND(OpenSSL_version);
+    XLIO_TLS_API_FIND(OpenSSL_version_num);
     if (s_tls_api.EVP_CIPHER_CTX_new && s_tls_api.EVP_CIPHER_CTX_free &&
         s_tls_api.EVP_CIPHER_CTX_reset && s_tls_api.EVP_aes_128_gcm && s_tls_api.EVP_aes_256_gcm &&
         s_tls_api.EVP_DecryptInit_ex && s_tls_api.EVP_DecryptUpdate &&
         s_tls_api.EVP_CIPHER_CTX_ctrl && s_tls_api.EVP_DecryptFinal_ex &&
         s_tls_api.EVP_EncryptInit_ex && s_tls_api.EVP_EncryptUpdate &&
         s_tls_api.EVP_EncryptFinal_ex) {
+        if (s_tls_api.OpenSSL_version) {
+            vlog_printf(VLOG_DEBUG, "Found OpenSSL version: %s\n", s_tls_api.OpenSSL_version(0));
```
BasharRadya, line 118: consider changing the log severity to info.
it will help customers to quickly identify if they are using a wrong OpenSSL version.

pasis: we should print minimum information on the screen and I think printing openssl version will be overhead, especially if the application is not TLS oriented. XLIO only needs the openssl symbols to function properly, so there is no wrong version: openssl-1.x can use KTLS (there is abandoned fork with KTLS support) or openssl without KTLS can be used intentionally and KTLS will never be requested, so extra information on screen can lead to confusion.

src/core/sock/sockinfo_ulp.cpp
pasis: Then numeric version branch is unlikely. Even openssl 1.x, bundled into WRK, provides the string API. OpenSSL in its history already changed the format of the version encoding. Besides, there is libressl with own versioning. So, I'd keep the number as is and even is we bump into such an old openssl somehow - we can decode the version manually. I'd prefer to keep the code more compact and not to decode the version.

BasharRadya: Okes

BasharRadya:
bot:retest

PR #349, opened 2025-05-07
Fix tls python offload with libxlio

BasharRadya:
## Description

##### What

##### Why ?

##### How ?
architecture, approach, etc._

## Change type
- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list


src/core/sock/sockinfo_ulp.cpp
```diff
@@ -508,6 +533,9 @@ int sockinfo_tcp_ops_tls::setsockopt(int __level, int __optname, const void *__o
             return -1;
         }
     } else {
+#ifdef DEFINED_UTLS
```
AlexanderGrissik, line 536: No need for this ifdef the whole code is under this ifdef

src/core/sock/sockinfo_ulp.cpp
```diff
@@ -116,10 +116,34 @@ template <typename T> static void dlsym_default(T &ptr, const char *name)
...
 
-#define XLIO_TLS_API_FIND(__name) dlsym_default(s_tls_api.__name, #__name);
+static void *openssl_handle = nullptr;
 
-void xlio_tls_api_setup()
+#define XLIO_TLS_API_FIND(__name) dlsym_handle(s_tls_api.__name, #__name, openssl_handle);
+
+inline bool check_openssl_loaded()
+{
+    openssl_handle = dlopen("libssl.so.3", RTLD_NOW | RTLD_GLOBAL);
```
AlexanderGrissik, line 125: The name of the library can change in the future. at least it should be some changeable define somewhere

src/core/sock/sockinfo_ulp.cpp
```diff
@@ -116,10 +116,34 @@ template <typename T> static void dlsym_default(T &ptr, const char *name)
...
+        // OpenSSL symbols is already loaded
+        return;
+    }
+
+    if (!check_openssl_loaded()) {
+        vlog_printf(VLOG_DEBUG, "OpenSSL library not found or failed to load");
+        return;
+    }
+
     XLIO_TLS_API_FIND(EVP_CIPHER_CTX_new);
```
AlexanderGrissik, line 147: I suggest trying the symbols first the old way without openssl_handle as this is a more generic way and does not require loading a specific lib file. And only if it fails then to fallback to dlopen

src/core/sock/sockinfo_ulp.cpp
```diff
@@ -116,10 +116,34 @@ template <typename T> static void dlsym_default(T &ptr, const char *name)
...
+        return false;
+    } else {
+        vlog_printf(VLOG_DEBUG, "Successfully loaded libssl.so.3");
+        return true;
+    }
+}
+
+inline void xlio_tls_api_setup()
 {
+    if (openssl_handle) {
```
AlexanderGrissik, line 137: I suggest trying the symbols first the old way without openssl_handle as this is a more generic way and does not require loading a specific lib file. And only if it fails then to fallback to dlopen

AlexanderGrissik:
The changes are OK to be provided as temporary patch

src/core/sock/sockinfo_ulp.cpp
pasis: Agree, the application can be statically linked.

src/core/sock/sockinfo_ulp.cpp
```diff
@@ -508,6 +533,9 @@ int sockinfo_tcp_ops_tls::setsockopt(int __level, int __optname, const void *__o
             return -1;
         }
     } else {
+#ifdef DEFINED_UTLS
+        xlio_tls_api_setup();
```
pasis, line 537: Current implementation is not thread safe. Such a runtime initialization needs to be protected. consider some kind of thread safe singleton assuming that a pointer assignment is an atomic operation on the target platforms (no need to lock the check for openssl API initialization)

BasharRadya:
bot:retest

BasharRadya:
bot:retest

PR #356, opened 2025-05-19
issue: 4415379 TLS RX offload encryption fix and improve statistics

AlexanderGrissik:
## Description
TLS RX offload fallback functionality and statistics.
Fix TLS RX connection termination on SW fallback.

##### What
1. Fix TLS RX fallback encrypt path
2. Add socket statistics for all TLS RX fallback cases
3. Add general ring statistics for TLS RX/TX
4. Cache align statistics

##### Why ?
Functionality fix.

## Change type
- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list


src/core/util/xlio_stats.h
```diff
@@ -337,30 +345,38 @@ static const char *const ring_type_str[] = {"RING_ETH", "RING_TAP"};
...
+    uint64_t n_rx_interrupt_received;
+    uint32_t n_rx_cq_moderation_count;
+    uint32_t n_rx_cq_moderation_period;
+    uint32_t n_rx_tls_contexts;
+    uint32_t n_rx_tls_resyncs;
+    uint32_t n_rx_tls_auth_fail;
+    ring_type_t n_type;
+    void *p_ring_master;
+
+    // First cache line
```
pasis, line 358: I'd suggest to add a compile time assert to guarantee the cache line alignment (by field offset or sizeof of the inner struct)

src/core/util/xlio_stats.h
```diff
@@ -86,9 +86,10 @@ extern user_params_t user_params;
 
 // Epoll group stats
 typedef struct {
-    bool enabled;
-    int epfd;
     iomux_func_stats_t stats;
+    int epfd;
+    bool enabled;
+    uint64_t unused_align64[3];
```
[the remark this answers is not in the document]
AlexanderGrissik: I couldn't find a builtin way to force struct size. i added aligas(64) on the relevant structs but this only guarantees the start address and not the size. It should be find for simple arays but it will not work in case of malloc(N * sizeof(struct)) or similar. Also i added static_asserts on the size of relevant structs to be on the safe side.

src/core/util/xlio_stats.h
AlexanderGrissik: I added static assert outside the class on the offset as it does not recognize the class from within.

src/core/util/xlio_stats.h
```diff
@@ -464,7 +489,7 @@ typedef struct sh_mem_t {
      * (possibly recursively), may not be a member of a structure or an element of an array.
      * (However, these uses are permitted by GCC as extensions.)
      */
-    socket_instance_block_t skt_inst_arr[1]; // sockets statistics array
+    alignas(64) socket_instance_block_t skt_inst_arr[1]; // sockets statistics array
```
pasis, line 492: Open question:  where should we place alignas() specifier - at the beginning or at the end? We don't have explicit convention for that.

AlexanderGrissik: alignas replaced with simple paddings because of over alignment nuances

src/core/util/utils.h
```diff
@@ -12,18 +12,32 @@
...
+#include <limits.h>
 #include <exception>
+#include <cstdint>
+#include <cstddef>
 #include "vtypes.h"
 #include "utils/rdtsc.h"
 #include "vlogger/vlogger.h"
 #include "core/proto/mem_buf_desc.h"
-#include "core/util/xlio_stats.h"
+//#include "core/util/xlio_stats.h"
```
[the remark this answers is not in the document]
AlexanderGrissik: Fixed

AlexanderGrissik:
bot:retest

PR #415, opened 2025-09-01
issue: 4315534 Threads mode TX path

pasis:
## Description
Threads mode TX path and related changes.

##### What
Threads mode TX path and related changes.

##### Why ?
TX path implementation

## Change type
- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list

### DIFF
--- a/src/core/dev/ring_simple.cpp
+++ b/src/core/dev/ring_simple.cpp
@@ -17,7 +17,7 @@
 #define MODULE_HDR MODULE_NAME "%d:%s() "
 
 #define ALIGN_WR_DOWN(_num_wr_) (std::max(32, ((_num_wr_) & ~(0xf))))
-#define RING_TX_BUFS_COMPENSATE 256U
+#define RING_TX_BUFS_COMPENSATE 1024U
 
 #define RING_LOCK_AND_RUN(__lock__, __func_and_params__)                                           \
     __lock__.lock();                                                                               \
@@ -866,6 +866,7 @@ void ring_simple::return_tx_pool_to_global_pool()
     return_to_global_pool();
 }
 
+// Returns number of freed buffers.
 int ring_simple::put_tx_buffer_helper(mem_buf_desc_t *buff)
 {
     if (buff->tx.dev_mem_length) {
@@ -885,7 +886,6 @@ int ring_simple::put_tx_buffer_helper(mem_buf_desc_t *buff)
         buff->p_next_desc = nullptr;
         free_lwip_pbuf(&buff->lwip_pbuf);
         pool.push_back(buff);
-        // Return number of freed buffers
         return 1;
     }
     return 0;
--- a/src/core/event/entity_context.cpp
+++ b/src/core/event/entity_context.cpp
@@ -46,7 +46,7 @@
 
 entity_context::entity_context(size_t index)
     : poll_group(xlio_poll_group_attr {XLIO_GROUP_FLAG_SAFE | XLIO_GROUP_FLAG_DIRTY, nullptr,
-                                       nullptr, nullptr, nullptr})
+                                       entity_context_comp_cb, nullptr, nullptr})
     , m_index(index)
 {
     ctx_logdbg("Entity Context created");
@@ -71,8 +71,7 @@ void entity_context::process()
             listen_socket_job(job);
             break;
         case JOB_TYPE_SOCK_TX:
-            // Handle socket transmit job
-            // TODO: implement transmit logic
+            tx_data_job(job);
             break;
         case JOB_TYPE_SOCK_RX_DATA_RECVD:
             rx_data_recvd_job(job);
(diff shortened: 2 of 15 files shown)


src/core/dev/ring_simple.cpp
```diff
@@ -17,7 +17,7 @@
 #define MODULE_HDR MODULE_NAME "%d:%s() "
 
 #define ALIGN_WR_DOWN(_num_wr_) (std::max(32, ((_num_wr_) & ~(0xf))))
-#define RING_TX_BUFS_COMPENSATE 256U
+#define RING_TX_BUFS_COMPENSATE 1024U
```
AlexanderGrissik, line 20: Should we add conf param instead?

src/core/sock/sockinfo_tcp.cpp
```diff
@@ -5790,22 +5777,13 @@ void tcp_timers_collection::handle_timer_expired(void *user_data)
...
-         */
+        // TODO Trylock can miss a timer tick and we don't trigger it in unlock() anymore.
         if (!p_sock->trylock_tcp_con()) {
-            bool destroyable = false;
-            if (!p_sock->is_cleaned()) {
-                p_sock->handle_timer_expired();
-                destroyable = p_sock->is_destroyable_no_lock();
-            }
+            p_sock->handle_timer_expired();
+            bool destroyable = p_sock->is_destroyable_no_lock();
```
AlexanderGrissik, line 5922: AFAIR, we call unregister_socket_timer_and_delete to both remove the timer and the delete the socket. So we may run tcp timer for a socket that is pending deletion.

AlexanderGrissik: We may add is_cleaned check in handle_timer_expired tor resolve this.

src/core/event/entity_context.cpp
```diff
@@ -167,3 +175,21 @@ void entity_context::close_socket_job(const job_desc &job)
...
+{
+    mem_buf_desc_t *buf_list = reinterpret_cast<mem_buf_desc_t *>(userdata_op);
+
+    NOT_IN_USE(sock);
+    NOT_IN_USE(userdata_sq);
+
+    if (likely(buf_list->p_desc_owner)) {
+        buf_list->p_desc_owner->mem_buf_tx_release(buf_list, true);
+    } else {
+        vlog_printf(VLOG_WARNING,
```
AlexanderGrissik, line 191: Can we really have this situation?

src/core/sock/sockinfo_tcp.cpp
```diff
@@ -428,6 +428,13 @@ void sockinfo_tcp::set_xlio_socket(const struct xlio_socket_attr *attr)
 void sockinfo_tcp::set_entity_context(entity_context *ctx)
 {
     m_entity_context = ctx;
+    // To reuse Ultra API path of TX completions
+    m_p_group = ctx;
```
AlexanderGrissik, line 434: This breaks our other logics where we check for m_p_group and then for m_entity_context to decide if the socket is ultra api or entity context

src/core/sock/sockinfo_tcp.cpp
```diff
@@ -428,6 +428,13 @@ void sockinfo_tcp::set_xlio_socket(const struct xlio_socket_attr *attr)
...
+
+    assert(m_ops == m_ops_tcp);
+    m_ops_tcp = new sockinfo_tcp_ops_thread(this);
```
AlexanderGrissik, line 437: should we move here rx path and maybe other stuff as well?

pasis: We can consider this. Currently RX path is done via lwip callback. With full access to sockinfo_tcp. Moving this logic to ops::recv() can limit us. Anyway, ops functionality can be extended if needed.

src/core/dev/ring_simple.cpp
pasis: Not in current PR. Future PR will address buffer management and will use 256KB TX buffers, so original 256 hardcoded value should be ok there. I will remove this commit and address it in other PR.

src/core/event/entity_context.cpp
AlexanderGrissik: i guess yes. Without the check the code clears that there should always be an owner otherwise its a bug

src/core/sock/sockinfo_tcp.cpp
```diff
@@ -1154,34 +1138,12 @@ bool sockinfo_tcp::prepare_dst_to_send(bool is_accepted_socket /* = false */)
...
+    if (ret_val && m_p_group) {
         add_tx_ring_to_group();
-    } else if (ret_val && m_entity_context) {
-        // Add the rings of the incoming socket to the entity context.
-        add_tx_ring_to_entity_context();
     }
     return ret_val;
 }
 
-bool sockinfo_tcp::prepare_dst_to_send_entity_context()
```
AlexanderGrissik, line 1161: We should remove then dst:_entry::prepare_to_send_entity_context as well

pasis: Removed and fixed ring resolution with subsequent run of prepare_to_send()

PR #527, opened 2025-12-24
issue: 4808527 Replace malloc() with mmap() in xlio_allocator

pasis:
## Description
malloc() doesn't guarantee alignment. Although it uses mmap() internally for
big allocations, the heap allocator can be replaced.

madvise(MADV_DONTFORK) fails if the pointer is not aligned to the page size.
Unaligned pointer can also fail ibv_regmr() if it calls madvise() internally.

To eliminate possibility of the above problem, avoid using malloc() for the
registered memory and use either mmap() or posix_memalign().

Additionally, in the posix_memalign() method, align the allocation size to
the alignment value to prepare for possible future aligned_alloc()
integration.

##### What

- Replace malloc() with mmap() in xlio_allocator.
- Document page-alignment requirement for the external allocator in Ultra API.

##### Why ?
Avoid (unlikely) unaligned memory for XLIO buffers. Otherwise, madvise() fails.

## Change type
- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [x] Documentation content changes
- [ ] Tests
- [ ] Other

## Check list

### DIFF
--- a/src/core/dev/allocator.cpp
+++ b/src/core/dev/allocator.cpp
@@ -79,13 +79,13 @@ void *xlio_allocator::alloc(size_t size)
         }
         // Fallthrough
     case ALLOC_TYPE_ANON:
-        long page_size;
-        page_size = sysconf(_SC_PAGESIZE);
-        if (page_size > 0) {
-            m_data = alloc_posix_memalign(size, (size_t)page_size);
-        }
+        m_data = alloc_mmap(size);
         if (!m_data) {
-            m_data = alloc_malloc(size);
+            long page_size = sysconf(_SC_PAGESIZE);
+            if (page_size <= 0) {
+                page_size = 4096;
+            }
+            m_data = alloc_posix_memalign(size, (size_t)page_size);
         }
         break;
     case ALLOC_TYPE_EXTERNAL:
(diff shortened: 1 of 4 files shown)


src/core/dev/allocator.cpp
```diff
@@ -79,13 +79,13 @@ void *xlio_allocator::alloc(size_t size)
...
     case ALLOC_TYPE_ANON:
-        long page_size;
-        page_size = sysconf(_SC_PAGESIZE);
-        if (page_size > 0) {
-            m_data = alloc_posix_memalign(size, (size_t)page_size);
-        }
+        m_data = alloc_mmap(size);
         if (!m_data) {
-            m_data = alloc_malloc(size);
+            long page_size = sysconf(_SC_PAGESIZE);
```
tomerdbz: consider caching this value as a class member as you repeat this call.
btw - this is what xlio_heap::initialize does. maybe we can just get its cached value?

src/core/dev/allocator.cpp
```diff
@@ -188,6 +200,11 @@ void xlio_allocator::dealloc()
     case ALLOC_TYPE_ANON:
         free(m_data);
         break;
+    case ALLOC_TYPE_MMAP:
+        if (munmap(m_data, m_size) != 0) {
+            __log_info_dbg("munmap failed (errno=%d)", errno);
```
tomerdbz: I'd argue it's not debug - allocation failures will mean something bad sooner or later

tomerdbz (changes requested):
Great work Dima! :) Small changes needed, we're almost there!

PR #561, opened 2026-01-27
issue: 4684689 Improve JSON schema documentation

tomerdbz:
## Description
Add comprehensive documentation to xlio_config_schema.json for multiple configuration parameters including detailed explanations of:

- TX/RX prefetch bytes: CPU cache mechanics, value ranges, tuning recommendations
- RX buffer size: Striding RQ interaction, LRO impact, memory calculations
- TCP segments batch sizes: Three-level caching architecture explanation
- GRO streams: Aggregation mechanics, memory overhead, monitoring tips
- Override rcvbuf limit: UDP socket buffer behavior, performance tradeoffs
- Nginx parameters: Flow steering mechanics, worker distribution, socket pooling
- Profile spec: Complete documentation for all predefined profiles

Documentation includes architecture-specific behavior, performance impact tables, interaction with other parameters, and actionable tuning recommendations.

Also adds missing "minimum" constraints for several integer parameters.

##### What
Improve JSON schema documentation for configuration parameters

##### Why ?
Improve Context for devs and AI tools

##### How ?
architecture, approach, etc._

## Change type
- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list


docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3622 @@
+# XLIO Configuration Reference
+
+This file documents all 124 XLIO runtime configuration parameters with their types, defaults, environment variables, and constraints.
+
+> **Auto-generated** from the JSON schema by `generate_docs.py`. Do not edit manually.
+
+## ACCELERATION_CONTROL
```
pasis, line 7: [comment] With current ordering there is a high chance that users see less frequently used or expert level parameter during reading the document. Therefore, it's suitable for workflow when users search for specific parameter or keywords. Since the parameter name can be unknown to user and they search by desired behavior, maybe we can simplify the search by adding keywords or something else.
No action needed.

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3622 @@
...
+
+### `acceleration_control.app_id`
+
+> **Type:** string
+
+Maps to **XLIO_APPLICATION_ID** environment variable.
```
pasis: [suggestion] Mapping to the old config name visually stick to the description. Can it be similar to "Default:" or "Type:"? Especially without extra new line all the description is displayed as a single paragraph.

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3622 @@
...
+---
+
+## APPLICATIONS
+
+### `applications.nginx.distribute_cq`
+
+> **Type:** boolean
+
+Maps to **XLIO_DISTRIBUTE_CQ** environment variable.
+Distributes completion queue interrupts across CPU cores based on worker ID.
```
pasis: s/worker ID/NGINX worker index/

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3622 @@
...
+
+**Completion Vectors:** Each completion queue maps to an MSI-X interrupt line via a
+completion vector index. The NIC's num_comp_vectors typically equals available CPU cores.
```
pasis: This is not user friendly technical details.

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3622 @@
...
+
+**Behavior:**
+- Enabled (worker process): comp_vector = worker_id % num_comp_vectors
+- Disabled: All completion queues use comp_vector=0 (all interrupts to CPU 0)
```
pasis: CPU 0 is incorrect information. It uses completion channel 0 which can have any CPU affinity.

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3622 @@
...
+> **Type:** boolean
+
+Maps to **XLIO_DISTRIBUTE_CQ** environment variable.
+Distributes completion queue interrupts across CPU cores based on worker ID.
+
+**Completion Vectors:** Each completion queue maps to an MSI-X interrupt line via a
+completion vector index. The NIC's num_comp_vectors typically equals available CPU cores.
+
+**Behavior:**
+- Enabled (worker process): comp_vector = worker_id % num_comp_vectors
```
pasis: [suggestion]
```
Enabled: completion queues are distributed across completion channels in round-robin manner.
```
`(worker process)` is confusing.
Formulas and variable names don't look user friendly.

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3622 @@
...
+- Disabled: All completion queues use comp_vector=0 (all interrupts to CPU 0)
+
+Only affects interrupt-driven I/O (epoll/blocking). No effect in pure polling mode.
+
+**Tradeoffs:**
+- *false* (default): Simple, predictable. Risk of CPU 0 hotspotting under high traffic
+  (connections per second or requests per second) with multiple workers.
+- *true*: Spreads interrupt load across cores. Better for multi-worker Nginx/Envoy with
+  high connection or request rates.
```
pasis: s/CPU 0/single CPU/

pasis: [comment] The text sounds severe, however, we have already targeted the root cause of high softIRQ cpu usage and don't keep cqfd in every socket's epfd. So, this option just distributes interrupts to multiple CPUs and probably can improve latency in the interrupt mode.
In the past, softIRQ could reach multiple CPU cores in specific scenarios and this is fixed.

pasis (commented):
Only several options are reviewed.

General comments:
* There are formulas and terms in a variable name style. Those terms are not defined and can be confusing for a user. I'd suggest to describe parameters with a human language instead of "machine" language. If the formula simplifies description or understanding, consider defining/describing each operand.
* I see XLIO short parameter names and non-defined variables used across the text. It's impossible to distinguish the parameter names unless user knows the name. I'd suggest visual identification of parameter names, for example, `param_name`. Would be perfect to represent them as internal links that lead to the parameter description (if this is feasible).
* [comment] tags in the review don't require any actions and just provide extra information.

docs/xlio_config_reference.md
tomerdbz: NGINX worker index implies this is ngx_worker, it's not. I changed though to worker process index. how's that?

docs/xlio_config_reference.md
tomerdbz: Cool - learned something new today.

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3622 @@
...
+  acceleration. Each connection uses one flow table entry.
+- *true:* One rule per local IP, unlimited connections.
+  Flow tags are masked (software classification, higher
+  per-packet CPU). Requires a unique local IP per XLIO
+  ring.
+
+**Sizing:** For TCP clients (outgoing connect()) where
+ephemeral ports would exhaust the hardware flow table.
+For servers (incoming accept()), prefer 3t_rules instead.
+Symptom of exhaustion: "attach_flow" failures in XLIO
```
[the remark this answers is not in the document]
tomerdbz: Fixed

generate_docs.py
```diff
@@ -1,363 +1,601 @@
...
+
+def _cli() -> None:
+    """CLI entry point with error handling."""
+    try:
+        main()
+    except (GenerationError, OSError) as exc:
+        sys.exit(f"Error: {exc}")
+    except (KeyError, TypeError, ValueError, IndexError) as exc:
+        log.debug("Schema structure error", exc_info=True)
+        sys.exit(f"Error: unexpected schema structure: {exc}")
```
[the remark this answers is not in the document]
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: consider removing the worker ID terminology ("distributes across cpu cores...")

README
```diff
@@ -161,1148 +161,13 @@ Example:
...
-      Example: profiles.spec=nginx_dpu applications.nginx.workers_num=<N>
+XLIO exposes runtime configuration parameters organized into hierarchical groups.
 
-   - "nvme_bf3" or 4
-      Optimized for SPDK solution over NVIDIA DPU BF3
-      Example: profiles.spec=nvme_bf3
-Default value is 0
+For the complete reference — including types, defaults, environment variables, and constraints — see:
 
+**[Configuration Reference](docs/xlio_config_reference.md)**
```
pasis: README file is not a markdown. We have options here:

1. Rewrite to plaintext with mentioning where to find the reference docs.
2. Convert the README to the markdown format and rename it (we already have README.md)
3. Do the item (1) and add the link to a proper place in the README.md, so users can quickly jump there from the main github page

README
```diff
@@ -161,1148 +161,13 @@ Example:
...
-   - "ultra-latency" or 1
-     Optimized for use cases that are keen on latency even more. This mode uses
-      single threaded model, avoids OS polling and progress engine.
-      Example: profiles.spec=ultra-latency
-
-   - "nginx" or 2
-      Optimized for nginx. This profile must be used to offload nginx. This profile
-      is turned indirectly by setting:
-      applications.nginx.workers_num=<N> where N is the number of nginx workers.
+# Configuration Values
```
pasis: This is not a markdown file. Follow the header style like in the rest of the document (unless it's converted to markdown).

README
pasis, line 57: Remove the list of parameters and keep only simple header example. Better to keep single point of knowledge for the params, otherwise, README version can be outdated since it's not autogenerated.

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+This file documents all 124 XLIO runtime configuration parameters with their types, defaults, environment variables, and constraints.
+
+> **Auto-generated** from the JSON schema by `generate_docs.py`. Do not edit manually.
+
+## ACCELERATION_CONTROL
+
+### `acceleration_control.app_id`
+
+> **Type:** string
+> **Maps to:** `XLIO_APPLICATION_ID`
```
pasis, line 13: Please print every tag (like Type, Maps to, etc) on a new line. It's not convenient when they're displayed in a single line and some information can be visually missed if user doesn't expect it (for example, potential 3rd optional tag at the end of the paragraph).

Relevant for the entire document

docs/xlio_config_reference.md
pasis: Lets remove this paragraph at all. The behavior description is enough for user to get all required information. Let's keep it simple.

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+**Completion Vectors:** Each completion queue maps to an MSI-X interrupt line via a
+completion vector index. The NIC's num_comp_vectors typically equals available CPU cores.
+
+**Behavior:**
+- Enabled: Completion queues are distributed across completion vectors in round-robin
+  fashion based on the worker process index.
+- Disabled: All completion queues share the same completion vector, typically concentrating
+  interrupt processing on a single CPU.
+
+Only affects interrupt-driven I/O (epoll/blocking). No effect in pure polling mode.
```
pasis: [suggestion] s/pure polling/busy polling/.

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+
+**Tradeoffs:**
+- *false* (default): All completion queues share a single completion vector. Simple and
+  predictable. Interrupt processing for all workers is handled by one CPU.
+- *true*: Distributes interrupts across CPUs, which can improve latency in interrupt-driven
+  mode by reducing per-CPU interrupt processing load.
+
+**Auto-enabled** when nginx profile is active ([`workers_num`](#applicationsnginxworkers_num) > 0 or spec=nginx).
```
pasis: Should `spec` be a link too?

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+
+Internally, XLIO rounds the worker count up to the nearest power of 2, then builds a bitmask from
+the rounded count and the stride. The bitmask intentionally excludes the least significant bit of
+the port number (see "Why Minimum is 2" below). Each worker is assigned a unique value within that
+mask, and the NIC routes packets whose source port matches the worker's value.
+
+**Why Minimum is 2:**
+The minimum value of 2 ensures the least significant bit of the source port is excluded from
+steering decisions. Client ephemeral source ports have no predictable pattern in their lowest bit,
+so including it would cause uneven distribution across workers.
```
pasis, line 98: [comment] Not sure this should be documented here: the main reason of ignoring the least significant bit is because `connect()` in Linux kernel allocated only even or odd (don't remember which exactly, bind() allocated the other half of ports respectively) port numbers util it consumes all available such ephemeral ports.

So we usually expect the least significant bit to be the same for most of the connections.

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+Divide expected concurrent UDP sockets by number of NGINX workers.
+For long-lived connections (seconds+): pool provides marginal benefit, use small values.
+For rapid socket cycling (sub-second lifetimes): size to expected concurrent count.
+
+*Example:*
+8 workers, expecting 1000 concurrent UDP upstream sockets with rapid turnover:
+~125 sockets/worker → pool_size of 100-150 captures most reuse benefit.
+
+*Memory cost:* pool_size × ~2-4KB per worker.
+(100 pool × 8 workers ≈ 1.6-3.2MB held in reserve)
```
pasis, line 181: Remove the example in brackets. The formula is enough for user to scale it to their requirements.

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+Pool size should match peak concurrent short-lived UDP sockets per worker.
+Divide expected concurrent UDP sockets by number of NGINX workers.
+For long-lived connections (seconds+): pool provides marginal benefit, use small values.
+For rapid socket cycling (sub-second lifetimes): size to expected concurrent count.
+
+*Example:*
+8 workers, expecting 1000 concurrent UDP upstream sockets with rapid turnover:
+~125 sockets/worker → pool_size of 100-150 captures most reuse benefit.
+
+*Memory cost:* pool_size × ~2-4KB per worker.
```
pasis, line 180: s/pool_size/udp_pool_size/

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+(100 pool × 8 workers ≈ 1.6-3.2MB held in reserve)
+
+**Related:** [`applications.nginx.udp_socket_pool_reuse`](#applicationsnginxudp_socket_pool_reuse) controls receive buffer recycling.
+
+**Default:** `0`
+
+### `applications.nginx.udp_socket_pool_reuse`
```
pasis, line 187: [comment] New config name sounds like it configures sockets in a poll instead of buffers batching strategy per socket.

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+*Non-zero values:*
+Lower values = more frequent reclamation, less memory per socket, more lock acquisitions.
+Higher values = better batching efficiency, more memory held per socket.
+
+*Memory formula:*
+Total pooled buffer memory ≈ [`udp_pool_size`](#applicationsnginxudp_pool_size) × this_value × buffer_size
+
+*Example:*
+[`udp_pool_size`](#applicationsnginxudp_pool_size)=50, `udp_socket_pool_reuse`=64, MTU=1500:
+50 × 64 × 1500 ≈ 4.8MB per worker in reserve.
```
pasis, line 214: Is it correct in case of striding RQ (default configuration)?

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+### `applications.nginx.workers_num`
+
+> **Type:** integer (min: 0)
+> **Maps to:** `XLIO_NGINX_WORKERS_NUM`
+
+Number of Nginx worker processes. **Must be > 0 to enable XLIO offloading for Nginx.**
+
+**Behavior:** When set to a value greater than 0, XLIO enables the nginx profile (ring per
+interface, TSO, 3-tuple rules), creates hardware flow steering rules per worker, and scales
+memory allocation.
```
pasis, line 228: Do we have NGINX profile description somewhere in the documents? If yes, better to point to the profile description instead of hardcoding what it does, because it can change in the future.

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+**Sizing:** Set to match your Nginx worker_processes configuration exactly.
+
+*Power-of-2 values (2, 4, 8, 16, ...):* Optimal. Equal traffic distribution per worker.
+
+*Non-power-of-2 values:* XLIO rounds up to the next power of 2 internally. The extra flow
+steering slots are distributed as secondary rules to the lowest-numbered workers, so those
+workers handle more traffic.
+
+**Memory:** 4GB per worker (16 workers or fewer) or 3GB per worker (more than 16 workers).
+Master process uses minimal memory.
```
pasis, line 239: Avoid memory description here - it's autotuning is overcomplicated and based on the target platform and number of workers.

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+
+**Example:** 3 workers (rounded up to 4 slots): Worker 0 gets 2 port slots (primary +
+secondary rule), Workers 1-2 get 1 slot each. Using 4 workers avoids this imbalance.
```
pasis, line 242: [comment] This information repeats across parameters. Would be nice to have section-level description with common knowledge. For example, `applications.nginx` group has a common description and then individual parameter without repeated common details.

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+> **Maps to:** `XLIO_SERVICE_NOTIFY_DIR`
+
+Directory where XLIO writes files for communication with xliod daemon.
+
+**Files created per process:**
+- `xlioagent.<pid>.sock` - Unix domain socket for state messages
+- `xlioagent.<pid>.pid` - PID file for termination monitoring
+
+**Critical:** This path must match the `--notify-dir` argument passed to xliod.
+Mismatched paths prevent daemon monitoring of XLIO processes.
```
pasis, line 263: [suggestion]
> This path must match the xliod `--notify-dir` argument if provided.

To avoid confusion that the parameter always needs to be configured.

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+Memory blocks are registered with RDMA hardware for DMA operations (RX/TX buffer pools).
+
+**Sizing:** XLIO requests exactly one block of this size from your allocator. If your
+allocator cannot provide it, or if the block is too small for XLIO's buffer pools,
+initialization fails with an exception. Set to your allocator's maximum available
+contiguous block, ensuring it meets XLIO's memory requirements.
+
+**Profile override:** nvme_bf3 profile sets this to 2GB for optimal SPDK performance.
+
+**Example:** DPDK/SPDK integration with 1GB hugepage pools: set to 1GB to match pool size.
```
pasis, line 390: Remove this example.

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+Faster startup. Risk: In cgroup-limited environments (containers, Kubernetes), mmap() may succeed
+even when hugepages exceed the limit. First memory access triggers SIGBUS with no earlier warning.
+
+**Decision guide:**
+- Containers, Kubernetes, cgroup-limited: Keep disabled (runtime SIGBUS is worse than startup delay)
+- Bare metal with verified hugepage config: Safe to enable for faster startup
+
+**Default:** `false`
+
+### `core.resources.external_memory_limit`
```
pasis, line 366: This is legacy parameter now and the behavior is to take maximum between core.resources.external_memory_limit and core.resources.memory_limit. It will be removed in the future.

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+*Sizing formula:*
+block_size ≈ (max_concurrent_buffers × 256 bytes) rounded up to reduce expansions.
+
+*Example:*
+Expecting 50,000 concurrent buffers per process: 50,000 × 256 = 12.8MB minimum.
+32MB default provides headroom for bursts without expansion.
+
+**Tradeoffs:**
+- *Higher values:* Fewer heap expansions (fewer mmap syscalls), more initial memory.
+- *Lower values:* Less memory waste, but more frequent expansions under load.
```
pasis, line 420: - metadata includes buffers descriptors and TCP segments objects
- buffer descriptor is 320 bytes AFAIR (unless it was optimized since then)
- the heap grows during warmup and after caching sufficient number of metadata objects we don't grow anymore, so the impact is only during warmup

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+
+**Profile override:** Nginx master processes automatically use 2MB (no traffic handling).
```
pasis, line 422: [comment] previously mentioned section/group level description of nginx details would be perfect place for such information. Master process reduces multiple resources usage, because traffic is not expected, however, some control flow packets are still possible (DNS, network user related services, etc).

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+
+**Default:** `32MB`
+
+### `core.resources.hugepages.enable`
+
+> **Type:** boolean
+> **Maps to:** `XLIO_MEM_ALLOC_TYPE`
+
+Controls whether XLIO uses huge pages (2MB/1GB) instead of regular 4KB pages for buffer
+allocation.
```
pasis, line 435: There can be various hugepages sizes (32KB, 512MB, etc). We support all of them. Remove the `2MB/1GB` part. Also remove explicit 4KB, because page size can be different and, for example, we support 64KB page size on Arm.

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+
+**Why it matters:** With 4KB pages and 2GB memory, the Translation Lookaside Buffer needs
+524,288 mappings. With 2MB hugepages, only 1,024. Each TLB miss costs ~100-200 CPU cycles
+(vs ~4 for hit), directly impacting packet latency.
```
pasis, line 439: Don't describe technical characteristics of HW (TLB cache miss cost). It's HW dependent. Just mention that this improves TLB cache footprint and NIC internal cache utilization.

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+
+**Behavior:**
+- true (default): Allocates via mmap() with MAP_HUGETLB. Falls back to malloc() if unavailable.
+- false: Uses standard malloc() or posix_memalign().
```
pasis, line 443: [suggestion]

> "Falls back to regular pages allocation if unavailable".

> "Uses regular pages for memory allocation"

today, posix_memalign() is used as first fallback method and malloc() after the it fails. In the near future, we will avoid malloc() at all and will use mmap() first and posix_memalign() if it fails.

False also forcibly disables hugepages for metadata allocation (this kind of allocations is allowed to use hugepages up to 32MB IIRC).

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+
+Pre-allocates memory for XLIO's hardware-registered buffer pools (RX/TX buffers, stride descriptors).
+
+**Behavior:** XLIO allocates a contiguous block at startup, registers it with RDMA hardware for
+DMA operations, and carves it into buffer pools. Memory is reserved upfront even when idle.
+
+**Sizing:** Match to peak buffer demand. Each ring consumes approximately:
+- With Striding RQ (default): ~2GB — the product of receive queue depth, strides per
+  element, and stride size (see hardware_features.striding_rq settings)
+- Without Striding RQ: ~32MB — the product of receive queue depth and receive buffer size
```
pasis, line 499: this calculation example doesn't take TX buffers into account

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+**Tradeoffs:**
+- *Higher values*: Handle traffic bursts without drops. Essential for multi-ring deployments.
+  Requires more hugepages, reserves memory even when idle.
+- *Lower values*: Smaller footprint, faster startup. Risk pool exhaustion under load
+  ("No buffers error" stat increases, packets dropped).
+
+**Monitoring:** Run `xlio_stats -p <pid> -v 3` (full view) to check buffer pool stats:
+- "No buffers error:" non-zero → increase `memory_limit` (pool exhaustion)
+- "Expands:" continuously increasing → increase `memory_limit` (reduce runtime growth overhead)
+- "SW RX Packets dropped:" non-zero → increase `memory_limit` or [`spare_buffers`](#performanceringsrxspare_buffers)
```
pasis, line 519: "SW RX Packets dropped" usually means that packet cannot be accepted (no socket found for example). So it should be removed from here. Insufficient memory can lead to HW drops, however, HW drops can happen because of other reasons too, so "No buffers error" is the best indication of insufficient memory.

pasis (commented):
Reviewed up to core.resources.memory_limit

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+
+**The Problem (multi-threaded apps only):**
+XLIO derives a hardware flow steering tag from the file descriptor number. Without deferral:
+1. Thread A calls close() on a socket, the kernel releases the file descriptor immediately
+2. Thread B creates a new socket and receives the same file descriptor number
+3. Hardware still has the old flow steering rule (being destroyed by Thread A)
+4. Packets for the new socket may be misrouted or dropped
+
+This race does NOT affect single-threaded apps or apps with long-lived sockets,
+because cleanup completes before any new socket is created.
```
pasis, line 638: The problem is relevant for single threaded too. Sockets are destroyed in the internal thread and destruction can be delayed. Also socket can become time-wait delaying the destruction by 2 minutes.

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+
+**Tradeoffs:**
+- *false* (default): Immediate fd release for faster resource recycling. Lower fd usage.
+  Safe for single-threaded apps or apps with long-lived sockets.
+- *true*: File descriptors remain open until destructor runs. Higher fd usage, may hit
+  fd limits sooner. Required for multi-threaded apps with rapid socket churn.
+
+**When to enable:** Multi-threaded applications creating/destroying sockets rapidly,
+or applications experiencing unexplained packet loss after socket close.
```
pasis, line 647: This param mitigates the following scenarios:

- sockfd reuse while we have leftover internal socket object alive (so we also reuse flow id for the new socket)
- port reuse while the leftover object is alive

This is likely not relevant for incoming TCP sockets. Full list of symptoms is hard to estimate. At least, early close() can lead to steering rule creation failure for the new socket.

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+  Required if the application may call fork().
+- *false*: Faster startup, no madvise() overhead. If fork() occurs, undefined behavior:
+  memory corruption, crashes, or hardware hangs.
+
+**Warning:** Many operations implicitly fork: system(), popen(), backtrace(), some logging
+libraries. Disable only with complete control over the application and guarantee of no fork().
+
+**Default:** `true`
+
+### `core.syscall.sendfile_cache_limit`
```
pasis, line 695: [comment] To me, sendfile_cache_limit is more like "resources" group than syscall.

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+**How it works:**
+Files sent via sendfile() are mmap'd, registered with RDMA hardware (ibv_reg_mr), and cached
+by file identity (device + inode). Different file descriptors to the same file share one mapping.
+When the cache exceeds this limit, LRU eviction removes oldest unused mappings.
+
+**Tradeoffs:**
+- *Higher values (default 10GB):* More files cached, better throughput for file-serving.
+  Memory is only consumed when files are actually mapped.
+- *Lower values:* Smaller footprint, but more frequent cache misses trigger mmap/registration
+  overhead (~10-100 microseconds per miss).
```
pasis, line 713: Remove the time estimation. We cannot guarantee these numbers.

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+
+**Note:** This cache is global across all sockets and threads.
+
+**Default:** `10GB`
+
+---
+
+## HARDWARE_FEATURES
+
+### `hardware_features.striding_rq.enable`
```
pasis, line 730: Striding RQ is also required to enable LRO

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+- Disabled: Lower latency, immediate per-packet delivery. Required for latency-sensitive
+  workloads (trading, real-time). Higher CPU at high packet rates.
+
+**Striding Receive Queue Interaction:**
+The maximum LRO payload is the smaller of the total stride buffer size (strides per element × stride size) and 64KB, rounded to 256 bytes.
+With defaults (2048 × 64 = 128KB), LRO can coalesce full 64KB segments.
+Lower stride configurations may limit aggregation size.
+
+**When to change:** Disable for latency-sensitive applications. Enable (or leave auto) for
+throughput-oriented workloads. Use auto for mixed workloads.
```
pasis, line 846: Remove "Use auto for mixed workloads". It makes impression that this is run-time changing parameter.

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+
+> **Type:** integer
+> **Maps to:** `XLIO_HIGH_WMARK_DEK_CACHE_SIZE`
+
+Maximum number of Data Encryption Key objects to cache for TLS TX offload.
+
+**How it works:** XLIO maintains a dual-cache for DEK reuse on the TX path. Closed connections
+return DEKs to the "put cache" (limited by this value). New connections draw from the
+"get cache". When the get cache empties, a crypto-sync operation swaps them. Creating new
+DEKs requires NIC firmware operations with latency cost; caching amortizes this.
```
pasis, line 861: dual-cache is the internal implementation details. Just mention that XLIO has a DEK cache to reduce NIC firmware operations.

Update: move this detail to the low watermark parameter if it cannot be described with general words.

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+
+**Sizing:** Match to peak concurrent TLS connections. The cache holds DEKs from closed
+connections for reuse by new ones. If connection churn exceeds cache size, new DEKs must
+be created (slower). For short-lived connections with high turnover, size to expected
+concurrent count. For long-lived connections, default is sufficient.
+
+**Tradeoffs:**
+- Higher values: Fewer crypto-sync operations, smoother latency. Higher NIC memory usage.
+- Lower values: Lower NIC resource consumption. More frequent crypto-sync with high churn.
```
pasis, line 870: "More frequent firmware operations". Let's avoid crypto-sync detail.

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+> **Type:** integer
+> **Maps to:** `XLIO_LOW_WMARK_DEK_CACHE_SIZE`
+
+Low watermark threshold that controls when to create new Data Encryption Keys versus
+recycling existing keys via crypto-sync.
+
+**Behavior:** When a TLS connection needs a key and the get cache is empty:
+- Put cache size <= min_size: Create new key (firmware cost, but one-time per key)
+- Put cache size > min_size: Crypto-sync swaps caches (firmware cost, reuses keys)
+
```
pasis, line 888: Try to avoid details about put and get lists. Think about mentioning that DEK objects need a batched FW operation before they can be reused and the parameter specifies the lower batch size limit (this is my understanding from the current description).

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+**Sizing:** Match to concurrent TLS connection count during peak traffic bursts.
+During connection spikes (new connections arriving faster than old ones close):
+- Value too low: Frequent crypto-sync operations cause latency spikes
```
pasis, line 891: s/crypto-sync/firmware/

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+> **Type:** boolean
+> **Maps to:** `XLIO_UTLS_RX`
+
+Offloads TLS decryption to the NIC's crypto engine using the Linux kTLS API.
+
+**Why default is false:** RX offload can perform worse than software decryption due to resync
+overhead. Out-of-order packets trigger resyncs that compete with TX traffic for Send Queue
+credits. Under load, resyncs get skipped and packets fall back to CPU decryption. High
+connection counts amplify this (each connection triggers independent resyncs). In extreme
+cases, throughput drops 20-50% compared to software-only.
```
pasis, line 914: Keep only the 1st sentence. Avoid implementation details. Add the following limitation to the 1st place: XLIO TLS RX offload applies system-wide limit to the number of TLS sessions which is below 64K sessions.

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+
+**Tradeoffs:**
+- false (default): Predictable performance, no NIC resources, works on any hardware.
+  Higher CPU usage, lower throughput ceiling.
+- true: Lower CPU, higher throughput under ideal conditions. Risk: degradation under load,
+  packet loss, or high connection count (thousands). Test with your workload before enabling.
```
pasis, line 920: There is no degradation under load. Packet loss and out of order trigger partial SW fallback or resyncs. There is no data about degradation from the number of connections (thousands), I'd say it's unexpected.

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+
+**Monitoring:** Check xlio_stats for "TLS Rx Resyncs" (should be low) and "TLS Rx fallback"
+counters. High values indicate RX offload may hurt performance.
+
+**Prerequisites:** ConnectX-6 DX or later, OpenSSL required, XLIO built with --enable-utls.
```
pasis, line 925: I'd say OpenSSL requirement needs to be described explicitly as a separate paragraph. XLIO requires the application to be linked against openssl and it searches for the openssl symbols in runtime to implement SW fallback which is essential part of the TLS RX offload feature.

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+- false: CPU handles encryption. Works on any hardware, simpler debugging, no NIC resources.
+  Higher CPU usage, lower throughput ceiling. Use when debugging TLS issues, hardware lacks
+  offload support, or very high connection count exhausts NIC Data Encryption Key cache.
+
+**Monitoring:** In xlio_stats, "TLS Tx Offload" confirms offload is active; "TLS Tx Resyncs"
+shows retransmission-triggered resyncs (unlike RX, TX resync has no performance impact).
+
+**Prerequisites:** ConnectX-6 DX or later, TLS library with kTLS support (OpenSSL 3.0+,
+GnuTLS 3.7+), XLIO built with --enable-utls. The TLS library handles kTLS setup internally
+(e.g., SSL_OP_ENABLE_KTLS in OpenSSL). No application code changes required.
```
pasis, line 951: [suggestion]

> ConnectX-6 DX or later, TLS library or application with kTLS support, XLIO built with --enable-utls.

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+- "disable" or 0: Forces TSO off
+- "enable" or 1: Forces TSO on if hardware supports it
+
+**Tradeoffs:**
+- Enabled: Higher throughput (line-rate on 100Gbps+), lower CPU, fewer work requests per byte.
+- Disabled: Required for network debugging (packet capture shows pre-segmentation sizes)
+  or virtualization with known TSO bugs.
+
+**Note:** TSO aggregation is limited by congestion window. During slow start with small cwnd,
+TSO cannot aggregate large segments regardless of max_size.
```
pasis, line 979: convert max_size to a link to be consistent with other params

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+
+**Values:**
+- "auto" or -1 (default): Enables based on adapter capability and ethtool setting
+- "disable" or 0: Forces TSO off
+- "enable" or 1: Forces TSO on if hardware supports it
+
+**Tradeoffs:**
+- Enabled: Higher throughput (line-rate on 100Gbps+), lower CPU, fewer work requests per byte.
+- Disabled: Required for network debugging (packet capture shows pre-segmentation sizes)
+  or virtualization with known TSO bugs.
```
pasis, line 976: remove "virtualization with known TSO bugs". Maybe there are some specific rare cases, but regular virtualization should be supported and widely used. Such comment can make impression that it should be disabled for any virtualized environment.

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+**Default:** `"auto" (-1)`
+
+### `hardware_features.tcp.tso.max_size`
+
+> **Type:** integer (bytes, min: 1) or string with size suffix (B, KB, MB, GB)
+> **Maps to:** `XLIO_MAX_TSO_SIZE`
+
+Maximum bytes aggregated into a single TSO super-packet before NIC segmentation.
+
+The actual payload per super-segment is the smallest of: the configured `max_size`,
```
pasis, line 993: I haven't met official term super-packet and super-segment. Unless this is well-known term, use "TSO segment" term.

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+hardware capability, and the TCP congestion window. The transmit buffer is also
+capped at the smaller of tx_buf_size and max_size.
```
pasis, line 995: The transmit buffer part is confusing. tx_buf_size param specifies TX buffer size which is used for copying userdata in the non-zerocopy path. Then multiple such buffers can be aggregated into a TSO segment with the HW limit of the scatter-gather elements per segment (what means number of TX buffers per segment).

Zerocopy path works differently and can put single big zerocopy chunk into TSO segment without splitting.

I'd suggest to remove the 2nd sentence (at least I don't understand it).

docs/xlio_config_reference.md
tomerdbz: fixed

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+
+**Sizing:** Default (256KB) handles most workloads. Increase for sustained bulk transfers.
+Decrease for latency-sensitive apps or to reduce burstiness on shared networks.
+
+**Tradeoffs:**
+- Higher: Fewer work requests per byte, higher throughput. Risk: bursty traffic.
+- Lower: Smoother traffic, faster first-byte. Risk: may not hit line-rate on 100Gbps+.
+
+**Hardware warning:** If NIC capability exceeds 256KB, XLIO logs a suggestion to increase max_size.
```
pasis, line 1004: Do we have such functionality? I'm not aware of it.

docs/xlio_config_reference.md
tomerdbz: Removed

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+
+**Supports suffixes:** B, KB, MB (e.g., "512KB", "1MB").
```
pasis, line 1006: AFAIK, all supported HW have limit of 256KB TSO size. Let's not exceed it in the examples.

pasis (commented):
Reviewed up to `monitor.exit_report`

README
tomerdbz: Done 3

README
tomerdbz: Fixed

README
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

tomerdbz:
ultra-api test has passed manually. waiting for green xlio-base.

I will address all of Dima's comments' post-merge

docs/xlio_config_reference.md
```diff
@@ -0,0 +1,3631 @@
...
+### `network.multicast.mc_loopback`
+
+> **Type:** boolean
+> **Maps to:** `XLIO_TX_MC_LOOPBACK`
+
+This parameter sets the initial value used by XLIO internally
+to control the multicast loopback packets behavior during transmission.
+An application that calls setsockopt() with IP_MULTICAST_LOOP will
+run over the initial value set by this parameter.
+Read more in 'Multicast loopback behavior' in notes section below.
```
pasis, line 1205: There is no `Multicast loopback behavior` section below

pasis (dismissed):
approving to unblock merging. The comments will be handled with additional effort.

[comment] Order of params is not prioritized.

pasis (approved):

docs/xlio_config_reference.md
tomerdbz: this will be removed as it's buggy
See https://github.com/Mellanox/libxlio/pull/563

tomerdbz: anyway - not sure when it will be merged so will fix - I read the kernel src and learned something new!

docs/xlio_config_reference.md
tomerdbz: I disagree - Without it, a user has to do the math themselves to decide "is this worth worrying about?"
I'd like it as a user. but if you insist, I'll remove

docs/xlio_config_reference.md
tomerdbz: I'll aggregate your parameter comments to a Task to consider changing the naming/namespaces

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: no, I changed the description completely. read now?

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: TODO - aggregate comment to ticket

docs/xlio_config_reference.md
tomerdbz: Thanks Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: where can I find this limit of 64K sessions?

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: See  lines 243-246 of ring_simple.cpp

docs/xlio_config_reference.md
tomerdbz: Fixed

docs/xlio_config_reference.md
tomerdbz: Fixed

PR #576, opened 2026-02-15
issue: 4684689 Replace exit_report with tuning report subsystem

tomerdbz:
## Description
Replace the old monitor.exit_report (vlog-only diagnostics at exit) with a unified tuning report that writes a structured text file. The report includes system context, effective config, runtime stats, socket summary, and performance indicators, and is intended to be pasted into an LLM or shared with engineers for tuning.

- Add monitor.report.mode (disable/auto/enable) and monitor.report.file_path with %d→PID substitution.
- In auto mode, generate report only when anomalies are detected (buffer allocation failures, HW RX drops).
- Keep config registry alive until after report generation; cache OFED version at init to avoid popen() during shutdown.
- Add tuning_report_printer module and gtest coverage for report creation, modes, path handling, and legacy env vars.

##### What
Replace exit_report with tuning report subsystem.

##### Why ?
OOB Experience

##### How ?
architecture, approach, etc._

## Change type
- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list


src/core/tuning_report_printer.cpp
```diff
@@ -0,0 +1,1085 @@
...
+bool tuning_report_has_errors()
+{
+    buffer_pool *pools[] = {g_buffer_pool_rx_rwqe, g_buffer_pool_rx_stride, g_buffer_pool_tx,
+                            g_buffer_pool_zc};
+    for (auto *pool : pools) {
+        if (pool) {
+            const bpool_stats_t *stats = pool->get_stats();
+            if (stats && stats->n_buffer_pool_no_bufs > 0) {
+                return true;
+            }
```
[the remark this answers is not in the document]
tomerdbz: g_buffer_pool_rx_ptr is not a distinct pool — it's an alias pointer that always points to either g_buffer_pool_rx_stride or g_buffer_pool_rx_rwqe depending on whether Striding RQ is enabled (see buffer_pool.cpp:19-24 and initialization in main.cpp:1136-1141). The four pools checked in tuning_report_has_errors() and log_summary_to_vlog() already cover the underlying object. Adding rx_ptr would double-count its allocation failures in the vlog summary. write_runtime_stats() includes it intentionally for diagnostic visibility (showing which pool the alias resolves to), but that's informational output, not an aggregation.

docs/xlio_tuning_report_reference.md
```diff
@@ -0,0 +1,1636 @@
...
+Even in full mode, if the stats pool is smaller than the total socket count,
+the report shows:
+```
+  # Note: per-socket traffic stats cover X/Y sockets (increase monitor.stats.fd_num for full coverage)
+```
+This means traffic split numbers (offloaded vs non-offloaded bytes) are
+**underestimates** — they only cover X of the Y sockets. Recommend increasing
+`monitor.stats.fd_num` to at least Y.
+
+### Truncated Reports
```
orshemesh-grok: I suggest to add to the report a final line saying "Report generated successfully", so the poor chatbot will not have to resort to such heuristics. These heuristics may change over time, and nobody will remember to update this document

docs/xlio_tuning_report_reference.md
```diff
@@ -0,0 +1,1636 @@
...
+**Check duration first:** Look at `Duration:` in the Preamble.
+- If Duration is in **milliseconds** (e.g., `Duration: 50ms`), the process exited
+  almost immediately. There is nothing meaningful to tune — counters are near zero
+  and throughput is meaningless. Ask the customer: "Your process ran for less than
+  a second. Please verify the application ran successfully and re-run for at least
+  5 minutes for meaningful diagnostics."
+- If Duration is **< 5 minutes**, throughput values include startup overhead. The
+  report will annotate this with `# Note: lifetime average, process ran < 5min`.
+
+**Determine report mode:**
```
orshemesh-grok: Suggest to put the report mode in the report itself, to avoid such heuristics

docs/xlio_tuning_report_reference.md
```diff
@@ -0,0 +1,1636 @@
...
+1. **MTU:** 1500 (default) — this is the primary bottleneck for bulk throughput.
+   Per-packet overhead at MTU 1500 is ~40x higher than at MTU 9000.
+2. **TSO/LRO:** At defaults (`auto`), should be active if NIC supports it. No
+   `ring_tx_tso_packets` visible — TSO may not be active.
+3. **Profile:** `none` — no profile active. For a bulk TCP RX workload, a profile
+   may help.
+4. **Memory:** Default 2 GB — no failures, but headroom for larger rings.
+5. **Ring sizes:** At defaults. With striding RQ enabled (default), RX ring is
+   only 128 elements — very small for 100 Gbps.
+6. **CPU:** Unknown — must ask.
```
orshemesh-grok, line 1045: Do we have a way allowing us to detect when customer application is simply bad?
Maybe it is doing a lot of CPU work or even, God forbid, sleeping?
Just a thought, this is not a CR comment.

docs/xlio_tuning_report_reference.md
```diff
@@ -0,0 +1,1636 @@
...
+| `ring_tls_*_resyncs` | `hardware_features.tcp.tls_offload.rx_enable`, `hardware_features.tcp.tls_offload.tx_enable` |
+| `ring_tls_rx_auth_fail` | `hardware_features.tcp.tls_offload.rx_enable` |
+| `non_offloaded_sockets` | `acceleration_control.default_acceleration`, `acceleration_control.rules` |
+| `total_rx_os_packets` / `total_tx_os_packets` | `acceleration_control.default_acceleration`, `acceleration_control.rules` (high OS packets = traffic bypassing XLIO) |
+| `poll_hit_rate` | `performance.polling.iomux.poll_usec`, `performance.polling.iomux.poll_os_ratio`, `profiles.spec` |
+| `listen_conn_dropped` | (application listen backlog — not an XLIO config) |
+| `strq_total_strides` | `hardware_features.striding_rq.enable`, `hardware_features.striding_rq.strides_num` |
+| `rx_throughput` / `tx_throughput` | `network.protocols.ip.mtu` (jumbo frames), `hardware_features.tcp.tso.enable`, `hardware_features.tcp.lro`, `profiles.spec`, `core.resources.memory_limit` |
+
+## Config Knob Quick Reference
```
orshemesh-grok, line 1326: Isn't this a repeat of the content of the README file?
What is the reason we need to repeat it (Instead of telling the chatbot to read the README file)?
I notice the validate_tuning_report_reference.py script will force us to notice this when things change, still I ask the question

src/core/dev/net_device_table_mgr.h
```diff
@@ -92,6 +94,25 @@ class net_device_table_mgr : public cache_table_mgr<int, net_device_val *>, publ
         m_closed_rings_rx_cq_drop_counter_lock.unlock();
     }
 
+    // Accumulate ring stats from a ring that is about to be destroyed.
```
orshemesh-grok, line 92: Suggest moving this function's body to the .cpp file

tests/validate_tuning_report_reference.py
```diff
@@ -0,0 +1,310 @@
+#!/usr/bin/env python3
+from __future__ import annotations
```
orshemesh-grok, line 2: Is there a unit test / gtest / CI test that runs this script ?
I think it is a must - otherwise people will forget to run it.

orshemesh-grok (commented):
This is amazing code !
I admire your vision and implementation !

src/core/tuning_report_printer.cpp
```diff
@@ -0,0 +1,1085 @@
...
+        return ma.size() > 0;
+    }
+
+    return false;
+}
+
+static void write_effective_config(FILE *f)
+{
+    fprintf(f, "## Effective Config (non-default only)\n");
+
```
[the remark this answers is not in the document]
tomerdbz: Checked the schema — the only array parameter (acceleration_control.rules) defaults to [], so the conservative path for same-size non-empty vectors never triggers on default values today. If a non-empty default array is added later, we can revisit. No change needed.

docs/xlio_tuning_report_reference.md
tomerdbz: the hueristic here is not great, I think I could change it w/o adding the report mode as it is very self-describing. (see the new guideline and let's discuss)

docs/xlio_tuning_report_reference.md
tomerdbz: The Quick Reference is a concise, troubleshooting-oriented subset. It exists so the doc is self-contained as LLM context.
take as an example:
The rule saying "increase core.resources.memory_limit" (line 431). a chatbot can't function around it w/o knowing the current default, valid range, or what the knob actually controls.

adding the entire README would pollute the context window, so I extracted only the must-haves

src/core/dev/net_device_table_mgr.h
tomerdbz: This follows the same pattern as increase_closed_rings_rx_cq_drop_counter five lines above.
Moving one without the other would be inconsistent, no?

can still move, but wondering what am I missing?

tests/validate_tuning_report_reference.py
tomerdbz: Before merging - I'll open a task for devops to add this to CI validation

docs/xlio_tuning_report_reference.md
tomerdbz: Agreed - fixed

docs/xlio_tuning_report_reference.md
tomerdbz: very interesting, yes - I think we could utlize monitor.stats.cpu_usage to get a good sense of what was done by the app.
this is extra effort - but I have also a cheap to implement idea even in this initial iteration - getrusage call. wdyt?

src/core/util/sys_vars.cpp
```diff
@@ -754,6 +754,9 @@ void mce_sys_var::get_env_params()
     memset(stats_filename, 0, sizeof(stats_filename));
     memset(stats_shmem_dirname, 0, sizeof(stats_shmem_dirname));
     memset(service_notify_dir, 0, sizeof(service_notify_dir));
+    memset(report_file_path, 0, sizeof(report_file_path));
+    memset(cached_ofed_version, 0, sizeof(cached_ofed_version));
+    memset(&init_time_monotonic, 0, sizeof(init_time_monotonic));
```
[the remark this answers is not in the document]
tomerdbz: Great catch!

docs/xlio_tuning_report_reference.md
tomerdbz: Fixed terminology

src/core/config/runtime_registry.cpp
```diff
@@ -159,7 +175,17 @@ void runtime_registry::runtime_char_array_t::set_value_from_any(const std::exper
...
+    // stats file, report file, app_id). Matches the legacy env-var path
+    // (mce_sys_var::read_env_variable_with_pid).
+    auto pos = value.find("%d");
+    if (pos == std::string::npos) {
+        strncpy(m_ptr, value.c_str(), m_size - 1);
+        return;
+    }
+    std::string resolved = value.substr(0, pos) + std::to_string(getpid()) + value.substr(pos + 2);
+    strncpy(m_ptr, resolved.c_str(), m_size - 1);
 }
```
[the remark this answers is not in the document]
tomerdbz: out of scope

src/core/dev/net_device_table_mgr.cpp
```diff
@@ -592,6 +616,29 @@ void net_device_table_mgr::new_link_event(const netlink_link_info *info)
      */
 }
 
+void net_device_table_mgr::increase_closed_rings_rx_cq_drop_counter(uint64_t count)
+{
+    m_closed_rings_rx_cq_drop_counter_lock.lock();
+    m_closed_rings_rx_cq_drop_counter += count;
+    m_closed_rings_rx_cq_drop_counter_lock.unlock();
+}
```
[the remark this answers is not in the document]
tomerdbz: out of scope

orshemesh-grok (approved):

tomerdbz:
ultra-api test passed locally. ready once xlio-base is green

tests/validate_tuning_report_reference.py
tomerdbz: https://jirasw.nvidia.com/browse/HPCINFRA-4275

PR #577, opened 2026-02-18
issue: 4892549 Fix UAF in ZCOPY TX callback during ring teardown

tomerdbz:
## Description
During poll_group destruction, sockets are closed before rings are released. When the ring tears down, it drains the TX completion queue and free_lwip_pbuf() fires the ZCOPY callback for outstanding buffers. This callback dereferences the socket (via tx.zc.ctx) which has already been freed, causing a use-after-free.

Suppress the ZCOPY completion callback during ring teardown by stripping the ZCOPY flag in put_tx_buffer_helper() when m_up_tx is false. The flag is already set to false in stop_active_queue_tx() before the CQ drain begins, so no new synchronization is needed.

##### What
Fix UAF in ZCOPY TX callback during ring teardown

##### Why ?

##### How ?
architecture, approach, etc._

## Change type
- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list

### DIFF
--- a/src/core/dev/ring_simple.cpp
+++ b/src/core/dev/ring_simple.cpp
@@ -846,6 +846,13 @@ int ring_simple::put_tx_buffer_helper(mem_buf_desc_t *buff)
     if (buff->lwip_pbuf.ref == 0) {
         descq_t &pool = buff->lwip_pbuf.type == PBUF_ZEROCOPY ? m_zc_pool : m_tx_pool;
         buff->p_next_desc = nullptr;
+
+        // Workaround for RM#4917604:
+        // During ring teardown, suppress ZCOPY completion callbacks to avoid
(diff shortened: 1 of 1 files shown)


tomerdbz:
@pasis - core issue is that xlio_perf didn't wait for termination event - will fix this as well for a good API example.

but, it shouldn't cause a UAF and possibly a crash - so the fix is a must imho

tomerdbz:
bot:retest

tomerdbz:
@galnoam - I think we should consider fixing this. see the updated ticket - the UAF still exists.
I believe this is the correct fix as well - during teardown the QP is in ERROR state. (modify_queue_to_error_state is called) These "completions" are not real application events, and we should make sure they are not called.

pasis (approved):

PR #580, opened 2026-03-01
issue: 4683517 Fallback to passthrough on entity-context connect failure

BasharRadya:
On connect run-time (e.g. tx/rx ring creation failure) failure in worker-thread mode, fall back to passthrough
(handle_close + SYSCALL(connect)) instead of failing the connection.

## Description

##### What

##### Why ?

##### How ?
architecture, approach, etc._

## Change type
- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list

### DIFF
--- a/contrib/jenkins_tests/gtest.sh
+++ b/contrib/jenkins_tests/gtest.sh
@@ -136,6 +136,42 @@ rc=$(($rc+$?))
 eval "${sudo_cmd} $timeout_exe env XLIO_WORKER_THREADS=3 GTEST_TAP=2 LD_PRELOAD=$gtest_lib $gtest_app $gtest_opt --gtest_filter=$worker_threads_filter --gtest_output=xml:${WORKSPACE}/${prefix}/test-worker-threads-not-pow2.xml"
 rc=$(($rc+$?))
 
+# Passthrough tests
+
+passthrough_filter="tcp_listen_connect_nb.server_client_nb"
+
+# Passthrough via non-offloadable NIC (loopback)
+
+# worker threads mode
+gtest_opt_lo="--addr=127.0.0.1,127.0.0.1"
+eval "${sudo_cmd} $timeout_exe env XLIO_WORKER_THREADS=1 GTEST_TAP=2 LD_PRELOAD=$gtest_lib $gtest_app $gtest_opt_lo --gtest_filter=$passthrough_filter --gtest_output=xml:${WORKSPACE}/${prefix}/test-passthrough-loopback-wt.xml"
+rc=$(($rc+$?))
+
+# R2C mode
+eval "${sudo_cmd} $timeout_exe env GTEST_TAP=2 LD_PRELOAD=$gtest_lib $gtest_app $gtest_opt_lo --gtest_filter=$passthrough_filter --gtest_output=xml:${WORKSPACE}/${prefix}/test-passthrough-loopback-r2c.xml"
+rc=$(($rc+$?))
+
+###### Passthrough via XLIO offload rules
+
+
+# Create a temporary libxlio.conf with passthrough rules and point XLIO_CONFIG_FILE to it.
+server_ip=$(ip -f inet addr show net2 | awk '/inet / {print $2}' | cut -d/ -f1)
+passthrough_conf=$(mktemp /tmp/libxlio-passthrough-XXXXXX.conf)
+printf "use os tcp_client %s:*:*:*\nuse os tcp_server %s:*\n" "$server_ip" "$server_ip" > "$passthrough_conf"
+
+# worker threads mode
+eval "${sudo_cmd} $timeout_exe env XLIO_CONFIG_FILE=$passthrough_conf XLIO_WORKER_THREADS=1 GTEST_TAP=2 LD_PRELOAD=$gtest_lib $gtest_app $gtest_opt --gtest_filter=$passthrough_filter --gtest_output=xml:${WORKSPACE}/${prefix}/test-passthrough-rules-wt.xml"
+rc=$(($rc+$?))
+
+# R2C mode
+eval "${sudo_cmd} $timeout_exe env XLIO_CONFIG_FILE=$passthrough_conf GTEST_TAP=2 LD_PRELOAD=$gtest_lib $gtest_app $gtest_opt --gtest_filter=$passthrough_filter --gtest_output=xml:${WORKSPACE}/${prefix}/test-passthrough-rules-r2c.xml"
+rc=$(($rc+$?))
+
+rm -f "$passthrough_conf"
+###### End of Passthrough via XLIO offload rules
+
+
+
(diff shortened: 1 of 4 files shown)


src/core/event/entity_context.cpp
```diff
@@ -125,6 +126,14 @@ void entity_context::connect_socket_job(const job_desc &job)
...
         add_socket(reinterpret_cast<sockinfo_tcp *>(sock));
         reinterpret_cast<sockinfo_tcp *>(sock)->connect_entity_context();
+        if (sock->isPassthrough()) {
+            int fd = sock->get_fd();
+            sock_addr peer =
+                sock->get_connected_addr(); /* copy before handle_close may destroy sock */
+            handle_close(fd, false, true);
+            SYSCALL(connect, fd, peer.get_p_sa(), peer.get_socklen());
+            return;
+        }
```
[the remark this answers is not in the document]
BasharRadya: fixed

src/core/event/entity_context.cpp
```diff
@@ -125,6 +126,14 @@ void entity_context::connect_socket_job(const job_desc &job)
         sock->set_entity_context(this);
         add_socket(reinterpret_cast<sockinfo_tcp *>(sock));
         reinterpret_cast<sockinfo_tcp *>(sock)->connect_entity_context();
+        if (sock->isPassthrough()) {
+            int fd = sock->get_fd();
+            sock_addr peer =
+                sock->get_connected_addr(); /* copy before handle_close may destroy sock */
+            handle_close(fd, false, true);
+            SYSCALL(connect, fd, peer.get_p_sa(), peer.get_socklen());
```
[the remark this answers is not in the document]
BasharRadya: EPOLLIN SO_ERROR
will keep the app updated of the socket errors

contrib/jenkins_tests/gtest.sh
```diff
@@ -136,6 +136,44 @@ rc=$(($rc+$?))
...
+# R2C mode
+eval "${sudo_cmd} $timeout_exe env GTEST_TAP=2 LD_PRELOAD=$gtest_lib $gtest_app $gtest_opt_lo --gtest_filter=$passthrough_filter --gtest_output=xml:${WORKSPACE}/${prefix}/test-passthrough-loopback-r2c.xml"
+rc=$(($rc+$?))
+
+# Passthrough via XLIO offload rules
+######
+
+# Append passthrough rules to /etc/libxlio.conf, run tests, then remove them.
+server_ip=$(ip -f inet addr show net2 | awk '/inet / {print $2}' | cut -d/ -f1)
+passthrough_rules="use os tcp_client ${server_ip}:*:*:*\nuse os tcp_server ${server_ip}:*"
```
pasis: Maybe just disable entire TCP protocol to avoid run-time IP detection what can fail in tricky system configurations and running on other servers?

contrib/jenkins_tests/gtest.sh
```diff
@@ -136,6 +136,44 @@ rc=$(($rc+$?))
...
+eval "${sudo_cmd} cp /etc/libxlio.conf /etc/libxlio.conf.bak"
```
pasis: /etc/libxlio.conf can be absent, this complicates restoring the initial state. Better to distribute libxlio.conf file as part of gtest dir and provide the file with the option. Especially, if you remove IP autodetection and the file can be static.
Also note that the old config approach will be removed in the future and the rule will need conversion to the new config param. In theory, this parameter can be inline with the new config subsystem and you don't need any files then.

src/core/event/entity_context.cpp
```diff
@@ -125,6 +126,14 @@ void entity_context::connect_socket_job(const job_desc &job)
         sock->set_entity_context(this);
         add_socket(reinterpret_cast<sockinfo_tcp *>(sock));
         reinterpret_cast<sockinfo_tcp *>(sock)->connect_entity_context();
+        if (sock->isPassthrough()) {
+            int fd = sock->get_fd();
+            sock_addr peer =
+                sock->get_connected_addr(); /* copy before handle_close may destroy sock */
```
pasis: put the comment above the line to not split it,

src/core/event/entity_context.cpp
```diff
@@ -125,6 +126,14 @@ void entity_context::connect_socket_job(const job_desc &job)
...
+            handle_close(fd, false, true);
```
pasis, line 133: [question] AFAIU, handle_close() will initiate job which can be executed with a delay. Can the incomplete XLIO socket object interfere with the kernel connect()? For example, previously executed bind().

src/core/sock/sockinfo.h
```diff
@@ -312,6 +312,7 @@ class sockinfo {
     bool has_epoll_context() { return (!!m_econtext); }
     bool get_rx_pkt_ready_list_count() const { return m_n_rx_pkt_ready_list_count; }
     int get_fd() const { return m_fd; };
+    const sock_addr &get_connected_addr() const { return m_connected; }
```
pasis: [suggestion] consider renaming to something with "peername", because we already have getpeername which copies m_connected. And you need the same idea, but without copy.

contrib/jenkins_tests/gtest.sh
BasharRadya: we have some optimization that if all TCP ruels are consistent to use OS we fallback earlier in socket creation, so we don't reach the flows in connect() and listen() that i really want to check for xlio offload rules..



"        case SOCK_STREAM: {
            transport = __xlio_match_by_program(PROTO_TCP, safe_mce_sys().app_id);
            if (transport == TRANS_OS) {
                fdcoll_logdbg("All TCP rules are consistent and instructing to use OS.");
                return -1;
            }"

contrib/jenkins_tests/gtest.sh
BasharRadya: 1) current new subsytem inline option partially supports XLIO offload rules, and can't apply the rules i am trying to apply here @tomerdbz  correct me if i am wrong.
2) Code Amended

src/core/event/entity_context.cpp
BasharRadya: Done

src/core/event/entity_context.cpp
BasharRadya: This question is legit even for the already existing logic of passthrough in connect().

from what i saw i don't see an issue. socket destruction should not have side effects to the socket fd in kernel

src/core/sock/sockinfo.h
BasharRadya: getpeername  wont work because socket is already set to passtrhough in this case so it will call syscall getpeername .

Changed name to peername

BasharRadya:
@dpressle  Please review scripts related to CI.

BasharRadya:
bot:retest

dpressle (approved):

pasis (approved):

PR #583, opened 2026-03-10
4394054 POC hybrid KTLS offload

orshemesh-grok:
## Description
Add configuration to XLIO to limit amount of hw-offloaded KTLS TX sessions

##### Why ?
When amount of HW sessions goes above a certain threshold, performance gets hit.

## Change type
- [ ] Bugfix
- [X] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list
- [X] Documentation has been updated (if necessary)

### DIFF
--- a/src/core/sock/sockinfo_tcp.cpp
+++ b/src/core/sock/sockinfo_tcp.cpp
@@ -19,6 +19,7 @@
 #include "util/instrumentation.h"
 #include "util/list.h"
 #include "util/agent.h"
+#include "util/xlio_stats.h"
 #include "event/event_handler_manager.h"
 #include "event/event_handler_manager_local.h"
 #include "event/poll_group.h"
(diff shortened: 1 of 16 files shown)


src/core/sock/sockinfo_ulp.h
```diff
@@ -12,6 +12,7 @@
 #include "proto/tls.h" /* xlio_tls_info */
 #include "lwip/err.h" /* err_t */
 
+#include <atomic> /* std::atomic */
```
pasis: not needed here

src/core/sock/sockinfo_tcp.cpp
```diff
@@ -4637,6 +4630,16 @@ int sockinfo_tcp::tcp_setsockopt(int __level, int __optname, __const void *__opt
                         ret = -1;
                         break;
                     }
+                    auto max_sessions = safe_mce_sys().utls_max_sessions;
+                    auto num_tls = g_global_stat_static.n_tcp_tls_sessions.load();
```
pasis, line 4634: this is ok for PoC, however, there is a race here in multi-threaded applications: we can read the same value in multiple threads and make decision to create TLS resources and the stat will be increased later for every such thread. (review the scenario of n_tcp_tls_sessions == max - 1 and multiple threads run setsockopt() concurrently).

Few extra sessions can be an issue if we preallocate resources, for example.

src/core/sock/sockinfo_ulp.h
orshemesh-grok: Fixed

PR #585, opened 2026-03-15
issue: 4896482 Optimize remove_from_all_epfds lock contention

BasharRadya:
On every close(), XLIO calls remove_from_all_epfds(fd), which walked all epfds and took each epfd_info lock to check if the fd needs removal. This caused lock contention when other threads held epfd_info locks.

Solution:
- Offloaded sockets: Use existing m_econtext for  epfd lookup
- Non-offloaded fds: Use sharded map for lookup Add a sharded map (1024 buckets) to track non-offloaded fd->epfd mappings. Each bucket has its own lock, reducing contention by distributing entries.
- Multi-epfd case: Fall back to walking all epfds (rare)

Changes:
- Add src/core/util/sharded_map.h: Generic sharded map template class
- epfd_info: Maintain sharded map on epoll_ctl ADD/DEL and epfd close
- close(): Update remove_from_all_epfds function

## Description

##### What

##### Why ?

##### How ?
architecture, approach, etc._

## Change type
- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list

### DIFF
--- a/src/core/iomux/epfd_info.cpp
+++ b/src/core/iomux/epfd_info.cpp
@@ -294,6 +294,7 @@ int epfd_info::add_fd(int fd, epoll_event *event)
     } else {
         fd_rec.offloaded_index = -1;
         m_fd_non_offloaded_map[fd] = fd_rec;
+        g_p_fd_collection->add_non_offloaded_fd(fd, this);
     }
 
     __log_func("fd %d added in epfd %d with events=%#x and data=%#x", fd, m_epfd, event->events,
@@ -439,6 +440,7 @@ int epfd_info::del_fd(int fd, bool passthrough)
             // This can happen after bind(), listen() or accept() calls.
             m_fd_non_offloaded_map[fd] = *fi;
             m_fd_non_offloaded_map[fd].offloaded_index = -1;
+            g_p_fd_collection->add_non_offloaded_fd(fd, this);
         }
 
         remove_socket_from_ready_list(temp_sock_fd_api);
@@ -466,6 +468,7 @@ int epfd_info::del_fd(int fd, bool passthrough)
         fd_info_map_t::iterator fd_iter = m_fd_non_offloaded_map.find(fd);
         if (fd_iter != m_fd_non_offloaded_map.end()) {
             m_fd_non_offloaded_map.erase(fd_iter);
+            g_p_fd_collection->remove_non_offloaded_fd(fd, this);
         }
     }
 
--- a/src/core/sock/fd_collection.cpp
+++ b/src/core/sock/fd_collection.cpp
@@ -29,6 +29,9 @@
 #define fdcoll_logdbg     __log_dbg
 #define fdcoll_logfunc    __log_func
 
+// Sentinel value indicating fd is mapped to multiple epfds (caller should walk all)
+#define MULTI_EPFD_SENTINEL reinterpret_cast<epfd_info *>(1)
+
 fd_collection *g_p_fd_collection = nullptr;
 
 fd_collection::fd_collection()
@@ -506,6 +509,7 @@ int fd_collection::del_epfd(int fd, bool b_cleanup /*=false*/)
 
 void fd_collection::remove_epfd_from_list(epfd_info *epfd)
 {
+    m_non_offloaded_epfd_map.remove_all_for_value(epfd);
     lock();
     m_epfd_lst.erase(epfd);
     unlock();
@@ -562,15 +566,53 @@ int fd_collection::del_socket(int fd, sockinfo **map_type)
     return -1;
 }
 
+void fd_collection::add_non_offloaded_fd(int fd, epfd_info *epfd)
+{
+    m_non_offloaded_epfd_map.add_or_replace(fd, epfd, MULTI_EPFD_SENTINEL);
+    fdcoll_logdbg("fd=%d epfd=%p add_non_offloaded_fd", fd, epfd);
+}
+
+void fd_collection::remove_non_offloaded_fd(int fd, epfd_info *epfd)
+{
+    m_non_offloaded_epfd_map.remove_if_equals(fd, epfd);
+}
+
 void fd_collection::remove_from_all_epfds(int fd, bool passthrough)
 {
+    fdcoll_logfunc("fd=%d passthrough=%d", fd, passthrough);
+
+    // Hold lock to prevent epfd destruction while we use it.
+    // del_epfd() takes this lock before destroying epfd_info.
     lock();
-    for (epfd_info *ep = m_epfd_lst.front(); ep; ep = m_epfd_lst.next(ep)) {
-        ep->fd_closed(fd, passthrough);
+
+    // Offloaded sockets: use m_econtext (limited to one epfd)
+    sockinfo *sockfd = get_sockfd(fd);
+    if (sockfd) {
+        epfd_info *ep = sockfd->get_epoll_context();
+        if (ep) {
+            ep->fd_closed(fd, passthrough);
+        }
+        unlock();
+        return;
+    }
+
+    // Non-offloaded fds: use sharded map
+    epfd_info *epfd = m_non_offloaded_epfd_map.remove_and_get(fd, nullptr);
+    if (!epfd) {
+        fdcoll_logdbg("fd=%d not in non_offloaded_epfd_map", fd);
+        unlock();
+        return;
+    }
+
+    if (epfd != MULTI_EPFD_SENTINEL) {
+        epfd->fd_closed(fd, passthrough);
+    } else {
+        // fd is in multiple epfds: walk all
+        for (epfd_info *ep = m_epfd_lst.front(); ep; ep = m_epfd_lst.next(ep)) {
+            ep->fd_closed(fd, passthrough);
+        }
     }
     unlock();
-
-    return;
 }
 
 #if defined(DEFINED_NGINX)
--- a/src/core/sock/fd_collection.h
+++ b/src/core/sock/fd_collection.h
@@ -17,6 +17,7 @@
 #include "sock/sockinfo.h"
 #include "iomux/epfd_info.h"
 #include "utils/lock_wrapper.h"
+#include <util/sharded_map.h>
 
 typedef xlio_list_t<sockinfo, sockinfo::pending_to_remove_node_offset> sockinfo_list_t;
 typedef xlio_list_t<epfd_info, epfd_info::epfd_info_node_offset> epfd_info_list_t;
@@ -157,6 +158,10 @@ public:
      */
     void statistics_print(int fd, vlog_levels_t log_level);
 
+    // Add/remove non-offloaded fd->epfd mapping (called from epfd_info)
+    void add_non_offloaded_fd(int fd, epfd_info *epfd);
+    void remove_non_offloaded_fd(int fd, epfd_info *epfd);
+
 #if defined(DEFINED_NGINX)
     bool pop_socket_pool(int &fd, bool &add_to_udp_pool, int type);
     void push_socket_pool(sockinfo *sockfd);
@@ -188,6 +193,9 @@ private:
     epfd_info_list_t m_epfd_lst;
     // Contains fds which are in closing process
     sockinfo_list_t m_pending_to_remove_lst;
+
+    // Sharded map for non-offloaded fd -> epfd reverse lookup.
+    sharded_map<int, epfd_info *> m_non_offloaded_epfd_map;
 
     const bool m_b_sysvar_offloaded_sockets;
 
--- a/src/core/util/sharded_map.h
+++ b/src/core/util/sharded_map.h
@@ -1,0 +1,129 @@
+/*
+ * SPDX-FileCopyrightText: NVIDIA CORPORATION & AFFILIATES
+ * Copyright (c) 2021-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: GPL-2.0-only or BSD-2-Clause
+ */
+
+#ifndef SHARDED_MAP_H
+#define SHARDED_MAP_H
+
+#include <unordered_map>
+#include <functional>
+#include <utils/lock_wrapper.h>
+
+/**
+ * A sharded map that distributes entries across multiple buckets to reduce lock contention.
+ * Each bucket has its own lock, so operations on different buckets can proceed in parallel.
+ *
+ * @tparam Key    The key type
+ * @tparam Value  The value type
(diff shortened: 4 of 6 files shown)


src/core/util/sharded_map.h
```diff
@@ -0,0 +1,130 @@
...
+        b.lock.lock();
+        auto it = b.map.find(key);
+        if (it != b.map.end() && it->second == value) {
+            b.map.erase(it);
+            b.lock.unlock();
+            return true;
+        }
+        b.lock.unlock();
+        return false;
+    }
```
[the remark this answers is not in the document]
BasharRadya: intended behavior

src/core/sock/fd_collection.h
```diff
@@ -189,6 +190,16 @@ class fd_collection : private lock_mutex_recursive {
...
     sockinfo_list_t m_pending_to_remove_lst;
 
+    // Sharded map for non-offloaded fd -> epfd reverse lookup.
+    sharded_map<int, epfd_info *> m_non_offloaded_epfd_map;
+
+public:
+    // Add/remove non-offloaded fd->epfd mapping (called from epfd_info)
+    void add_non_offloaded_fd(int fd, epfd_info *epfd);
+    void remove_non_offloaded_fd(int fd, epfd_info *epfd);
+    void remove_all_non_offloaded_fds_for_epfd(epfd_info *epfd);
```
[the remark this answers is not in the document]
BasharRadya: done

src/core/sock/fd_collection.cpp
```diff
@@ -562,15 +562,61 @@ int fd_collection::del_socket(int fd, sockinfo **map_type)
...
+    }
+
+    if (epfd != sharded_map<int, epfd_info *>::multi_value_sentinel()) {
+        epfd->fd_closed(fd, passthrough);
+    } else {
+        // socket is in multiple epfds: walk all
+        for (epfd_info *ep = m_epfd_lst.front(); ep; ep = m_epfd_lst.next(ep)) {
+            ep->fd_closed(fd, passthrough);
+        }
+    }
```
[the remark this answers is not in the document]
BasharRadya: removed confusing debugging.
sentinal removed in remove_and_get

src/core/util/sharded_map.h
```diff
@@ -0,0 +1,130 @@
...
+
+private:
+    struct bucket_t {
+        std::unordered_map<Key, Value> map;
+        lock_mutex_recursive lock;
+    };
+
+    bucket_t m_buckets[N];
+
+    static size_t bucket_index(const Key &key) { return std::hash<Key> {}(key) & (N - 1); }
```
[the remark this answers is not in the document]
BasharRadya: thee is a comment that say need power of 2

src/core/sock/fd_collection.cpp
```diff
@@ -562,15 +562,60 @@ int fd_collection::del_socket(int fd, sockinfo **map_type)
...
+
+    // Offloaded sockets: use m_econtext (limited to one epfd)
+    sockinfo *sockfd = get_sockfd(fd);
+    if (sockfd) {
+        epfd_info *ep = sockfd->get_epoll_context();
+        if (ep) {
+            ep->fd_closed(fd, passthrough);
+        }
+        unlock();
+        return;
```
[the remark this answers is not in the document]
BasharRadya: "        // Currently XLIO does not support more then 1 epfd listed"

not supported case

src/core/util/sharded_map.h
```diff
@@ -0,0 +1,130 @@
...
+ * Each bucket has its own lock, so operations on different buckets can proceed in parallel.
+ *
+ * @tparam Key    The key type
+ * @tparam Value  The value type (pointer type expected)
+ * @tparam N      Number of buckets (should be power of 2 for efficient hashing)
+ */
+template <typename Key, typename Value, size_t N = 1024> class sharded_map {
+public:
+    // Sentinel value indicating key is mapped to multiple values (caller should walk all)
+    static inline Value multi_value_sentinel() { return reinterpret_cast<Value>(1); }
```
orshemesh-grok: The fact that you need a sentinel to make up for the fact that the map cannot store multiple values, should not be related to this generic class. I believe you should remove this (Actually, move this to the fd_collection, which is the user of this template, and see below my suggestions for handling the fact that this behavior is now controlled from the outside.

src/core/util/sharded_map.h
```diff
@@ -0,0 +1,130 @@
...
+
+    sharded_map() = default;
+    ~sharded_map() = default;
+
+    /**
+     * Add a key->value mapping.
+     * If key already exists with different value, marks it as multi_value_sentinel.
+     * @return true if this is a new entry, false if key already existed
+     */
+    bool add(const Key &key, Value value)
```
orshemesh-grok: /** 
 * Adds a key->value mapping.
 * If key already exists with the same value, does nothing
 * If key already exists with a different value, replace its value with the specified value 'replace_with'
 * @return ...
 * /
bool add_or_replace(const Key& key, Value value, Value replace_with)

src/core/util/sharded_map.h
```diff
@@ -0,0 +1,130 @@
...
+        bucket_t &b = m_buckets[idx];
+
+        b.lock.lock();
+        auto it = b.map.find(key);
+        if (it == b.map.end()) {
+            b.map[key] = value;
+            b.lock.unlock();
+            return true;
+        } else if (it->second != value && it->second != multi_value_sentinel()) {
+            it->second = multi_value_sentinel();
```
orshemesh-grok: } else if (it->second != value && it->second != replace_with) {
  it->second = replace_with;
}

src/core/util/sharded_map.h
```diff
@@ -0,0 +1,130 @@
...
+    Value remove_and_get(const Key &key)
+    {
+        size_t idx = bucket_index(key);
+        bucket_t &b = m_buckets[idx];
+
+        b.lock.lock();
+        auto it = b.map.find(key);
+        if (it == b.map.end()) {
+            b.lock.unlock();
+            return nullptr;
```
orshemesh-grok: nullptr only works if Value is a pointer type.
I suggest adding a parameter 'default_value' or 'not_found_value' which is returned if key is not found.

src/core/util/sharded_map.h
```diff
@@ -0,0 +1,130 @@
...
+            }
+            b.lock.unlock();
+        }
+        return removed;
+    }
+
+private:
+    struct bucket_t {
+        std::unordered_map<Key, Value> map;
+        lock_mutex_recursive lock;
```
orshemesh-grok: I agree with greptile, type should be lock_mutex.
You never call any function when you have the lock in your hands, so no need to make it recursive,

src/core/util/sharded_map.h
orshemesh-grok: I think this does not belong here - this is application logic.
See my fix to add()

src/core/util/sharded_map.h
orshemesh-grok: `bool add_or_replace(const Key &key, Value value, Value replace_value)`

and then use `replace_value` instead of `multi_value_sentinel()`

src/core/util/sharded_map.h
```diff
@@ -0,0 +1,130 @@
...
+        b.lock.unlock();
+        return false;
+    }
+
+    /**
+     * Remove a specific key->value mapping.
+     * Only removes if the current value matches.
+     * @return true if entry was removed, false otherwise
+     */
+    bool remove(const Key &key, Value value)
```
orshemesh-grok: I believe this should be renamed remove_if_equals().
In the future, someone could add a real simple remove().

src/core/util/sharded_map.h
orshemesh-grok: nullptr works only when Value is a pointer type.
I suggest to add a parameter 'default_value' or 'not_found_value' and return it instead of nullptr.

src/core/util/sharded_map.h
orshemesh-grok: Limiting to N-1 entries with "& (N-1)" works only if N is a power of 2.
I suggest to should replace it with "% N".
N is known at compile time, and the compiler is smart to know that when N is a power of 2, this can be replaced with & for a faster operation.

(If the compiler was stupid, I would suggest to check in the ctor that N is a power of 2, and throw if it is now, but there is no need to do this. And maybe in the future someone will want a specific value which is not a power of 2, maybe to match the amount of some external resource or something like that)

src/core/iomux/epfd_info.cpp
```diff
@@ -737,6 +740,8 @@ void epfd_info::ring_wait_for_notification_and_process_element(void *pv_fd_ready
 void epfd_info::clean_obj()
 {
     if (g_p_fd_collection) {
+        // Remove all non-offloaded fd->epfd mappings for this epfd before destruction
```
orshemesh-grok: Why not put the content of remove_all_non_offloaded_fds_for_epfd() inside remove_epfd_from_list() ?
Is there any scenario in which someone would want to call remove_epfd_from_list() and not call remove_all_non_offloaded_fds_for_epfd() ?

src/core/sock/fd_collection.cpp
```diff
@@ -562,15 +562,59 @@ int fd_collection::del_socket(int fd, sockinfo **map_type)
     return -1;
 }
 
+void fd_collection::add_non_offloaded_fd(int fd, epfd_info *epfd)
+{
+    m_non_offloaded_epfd_map.add(fd, epfd);
```
orshemesh-grok: I suggest to capture the return value and report it in the fdcoll_logdbg() function

src/core/sock/fd_collection.cpp
```diff
@@ -562,15 +562,59 @@ int fd_collection::del_socket(int fd, sockinfo **map_type)
...
+    fdcoll_logdbg("fd=%d epfd=%p add_non_offloaded_fd", fd, epfd);
+}
+
+void fd_collection::remove_non_offloaded_fd(int fd, epfd_info *epfd)
+{
+    m_non_offloaded_epfd_map.remove(fd, epfd);
```
orshemesh-grok: 1. Suggest capture return value and report it with fdcoll_logdbg. 
2. Based on greptile comment below (Which was resolved), I understand that you had a log here and you removed it. I suggest that instead of removing it, make it so that people understand it is OK if log shows "removed=0" for FDs which were in more than one epfd:
```
if removed:
    fdcoll_logdbg("fd=%d epfd=%p remove_non_offloaded_fd returning true", fd, epfd);
else
    fdcoll_logdbg("fd=%d epfd=%p remove_non_offloaded_fd returning false, this is expected if FD was in multiple epfds.", fd, epfd);

```

src/core/sock/fd_collection.cpp
```diff
@@ -562,15 +562,59 @@ int fd_collection::del_socket(int fd, sockinfo **map_type)
...
+}
+
+void fd_collection::remove_all_non_offloaded_fds_for_epfd(epfd_info *epfd)
```
orshemesh-grok: Repeating myself :-), I think this should be a part of remove_epfd_from_list()

orshemesh-grok (changes requested):
Nice !
And good that you did not change any lock flow -> Zero deadlock danger :-)

src/core/sock/fd_collection.cpp
BasharRadya: when we create the package we remove debug logs which will leave the captured value as unused variable and fail CI.

src/core/sock/fd_collection.cpp
BasharRadya: 1) i wouldn't like to add if condition (branch have cost) jsut for logs.
2) when we create the package we remove debug logs which will leave the captured value as unused variable and fail CI.

src/core/util/sharded_map.h
BasharRadya: Done

src/core/util/sharded_map.h
BasharRadya: Done

src/core/util/sharded_map.h
BasharRadya: Done

src/core/util/sharded_map.h
BasharRadya: Done

src/core/util/sharded_map.h
BasharRadya: Done

src/core/util/sharded_map.h
BasharRadya: Done

src/core/util/sharded_map.h
BasharRadya: Done

src/core/util/sharded_map.h
BasharRadya: Done

src/core/util/sharded_map.h
BasharRadya: Done

src/core/util/sharded_map.h
BasharRadya: Done

src/core/sock/fd_collection.cpp
BasharRadya: Done

src/core/iomux/epfd_info.cpp
BasharRadya: Done

src/core/util/sharded_map.h
```diff
@@ -0,0 +1,129 @@
...
+            return true;
+        } else if (it->second != value && it->second != replace_with) {
+            it->second = replace_with;
+        }
+        b.lock.unlock();
+        return false;
+    }
+
+    /**
+     * Remove a specific key->value mapping.
```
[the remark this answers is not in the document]
BasharRadya: should not reach this

src/core/util/sharded_map.h
```diff
@@ -0,0 +1,129 @@
...
+            }
+            b.lock.unlock();
+        }
+        return removed;
+    }
+
+private:
+    struct bucket_t {
+        std::unordered_map<Key, Value> map;
+        lock_mutex lock;
```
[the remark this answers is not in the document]
BasharRadya: comment removed

BasharRadya:
bot:retest

orshemesh-grok (approved):
Looks great !

BasharRadya:
bot:retest

galnoam:
decided to go with another fix.

PR #589, opened 2026-03-19
issue: 4934380 Fix config reference accuracy

tomerdbz:
## Description
Correct technical explanations (stride minimum, deferred_close, memory_limit_user), remove hardware-dependent latency numbers, fix variable name references, add missing info (TLS RX session limit, LRO/Striding RQ dependency), and unify profile cross-references.

Sync schema JSON descriptions with markdown.

##### What
Config reference fixes

##### Why ?
Fix comments of https://github.com/Mellanox/libxlio/pull/561

##### How ?
architecture, approach, etc._

## Change type
- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list


tomerdbz:
@pasis - addressed all your comments from 
https://github.com/Mellanox/libxlio/pull/561

you can keep commenting there

docs/xlio_config_reference.md
pasis: Looks like a long description for a simple substitution feature. Also pre-fork/post-fork looks confusing to me. We should re-init XLIO in the fork() and the substitution is done during this process, right?

I believe the same substitution feature is relevant for most of the path parameters. Maybe it needs to be described in one place and all such params just mention the behavior with a short note with a link?

In general, current param description is displayed as a single paragraph - split it into multiple.

pasis, line 1261: put new line before example

pasis, line 1336: during display this text is merged with the "enable" item - put new line before.

pasis, line 1375: Put new line before.

Also this param should support %d substitution. Consider marking all the path parameters with some metadata (group, type, etc that indicates them as a path). This can help to do automated checks or changes.

pasis: Split into multiple paragraphs for readability

pasis: currently we have Reno, not New Reno.

s/LWIP/LwIP/

pasis, line 1500: put new line before

pasis: TIME_WAIT is 2x MSL. So, with hardcoded MSL value in XLIO, this is exactly 120 seconds.

Make the same order of true and false items here and in tradeoffs section.

pasis: use style for `limit` instead of upper case. Visually it sticks to MSS, so better highlight it with bold font.

pasis: I doubt XLIO does PMTUD. It just respects MSS option from the opposite side. But the both sides can be XLIO. I'd suggest to focus here on the problem that there is smaller MTU on the route what is not visible for the both ends.

pasis, line 1: There is broken text layout across the document. I could miss some places. Make sure the text is split into proper paragraphs and the "Example" and other tags are placed on a new line (need to wrap with empty lines). Otherwise, the text becomes less readable when everything is merged into a single paragraph.

Pay attention on the comments for the path params. They have common properties (e.g. %d) that can be described in some general section and such params can point to the common description with a link under some "note: " tag. Also consider (optional) marking all the path params explicitly in the json schema, so we can do automated checks and edits.

pasis, line 1596: I'd suggest to add a simple intuitive description: When > 0, this feature combines Nagle/Minshall's algorithm for data below the threshold with the TCP_NODELAY behavior for above the threshold. This feature can be beneficial for applications that send header and payload of a high level PDUs with separate send operations.

pasis: Just "one MSS". Value depends on MTU and other factors (remote side can advertise lower MSS or MSS can be configured by user).

pasis: why a deadlock is possible? All inflight segments are acknowledged eventually and Nagle will progress with 1 segment at a time in the worst case. Mentioning a possibility of a deadlock can make impression that Nagle must be disabled in any production, however, there are probable specific scenarios when Nagle can be beneficial.

pasis: what is the scenario for 200ms latency? The latency will be either RTT or delayed ACK+RTT in the worst case. XLIO has up to 100ms delayed ACK period by default and I expect kernel has shorter. Maybe better to describe that delayed ack and RTT can affect limited small writes?

pasis: AFAIR, it means PUSH flag for every TCP segment and HW keeps it only in the last segment on the wire after TSO segment splitting.

I'm afraid, "last segment of each write" can be unambiguous for user. Because write can mean write()/send() operation. In our case, write is posting a WQE which is translated into multiple TCP segments on the wire. This can be too complicated for a user level param description, so let's think how we can rephrase it to something like "is set on the last segment of each bulk of segments sent to wire".

pasis: PSH flag doesn't affect, no need to describe transmission. You can mention that the param doesn't affect TSO functionality.

pasis, line 1647: [comment] This param was added as workaround for kernel behavior in GRO: when we send TSO segment with the PSH, HW sends multiple segments without the flag and puts it only to the last segment. When kernel receives the segments, it feels like GRO keeps the segments without the flag and pushes the PSH segments to the TCP processing out of order. This behavior triggered fast retransmissions after 3 out of order segments.

The above behavior was observed long time ago, so it needs confirmation that kernel's GRO is still affected and whether this is expected/documented behavior.

The bottom line, this is workaround param rather than performance (however, maybe it can have performance implications)

pasis: Don't mention specific numbers, better to refer to the delayed ACK timeout.

pasis: XLIO implements RFC and sends ACK for every other (every 2nd) TCP segment. LRO/GRO segments are ACKed immediately as they aggregate multiple TCP segments. Maybe it worth mentioning.

[question] Does automated script replaces the default 100ms if it's changed in the schema? If not, remove the number, it's available by the link anyway.

pasis: I'd argue it's better for low-latency. Currently, timers are executed in the internal thread and can cause lock contention on the sockets. The most expected behavior of the TCP timer (when delated ACK is disabled in low-latency) is RTO. RTO is a disaster for low-latency and needs to be avoided, so low value can hurt and RTO should be too rare to benefit from it.

pasis, line 1759: Add new line before.

docs/xlio_config_reference.md
```diff
@@ -1539,7 +1559,8 @@ sensitive apps or high connection counts. Default 1MB suits most datacenter and
 
 **Override:** Per-socket via setsockopt(SO_SNDBUF).
```
pasis: Unless it's changed, XLIO repeats Linux kernel behavior and doubles requested SNDBUF value. However, kernel uses extra space for metadata and XLIO counts it only for the payload.

Also Ultra API ignores send buffer.

No need to mention: there are situations, when XLIO can queue a single zerocopy segment beyond the send buffer.

docs/xlio_config_reference.md
pasis: which pool in this case? tcp_buffer_batch is per socket AFAIU and ring has own batching and caching abilities, so we lock ring (to access its local pool), not the global buffer pool.

Lock contention can happen when ring allocation logic is not suitable for the scenario, or some performance hurting features are enabled in the internal thread. If sockets are not migrated and ring allocation logic is good, we don't expect lock contention.

pasis: I guess this describes RX buffers reclaim behavior only. I'd say caching without delayed reclaim, because the buffers are reclaimed after reaching the threshold.

TX buffers are always returned to the ring immediately.

pasis: Needs to think how idle sockets affect striding RQ. We don't return strides, right? In this case entire big buffer cannot be reused.

pasis: From the code it looks like the param affects the maximum LRO aggregation (needs confirmation).

pasis: Remove memory per ring calculation, it's more complex because of caching, queued RX data, etc. With XLIO_MEMORY_LIMIT feature, the RX buffers memory footprint is adjustable, also exit report can show actual memory footprint of all the RX buffers if needed.

pasis: Don't recommend to match MTU, since this is default behavior. Original purpose likely was to support LRO without striding RQ (LRO was implemented originally before Striding RQ).

Maybe we can force to use striding RQ for LRO and remove this param at all.

pasis: This section doesn't seem right to me. There is no single recommendation and 10-20 segments is not expected value. Remove section for now unless we have a good recommendation.

pasis: I'd avoid specific TCP segment struct size here, because it can be reworked. Also I'd expect the struct to fit into cache line boundaries. Better mention that with default value, expansion takes 2-4 MB. Also XLIO aims to use hugepages by default is available and utilizes the hugepages efficiently (not strictly 16K segments, but how many fits into allocated hugepages).

pasis: does it mean that the batch size lower than per socket batch request? If yes, it's better to handle on the config level and make the current param max between configured value and per socket batch.

The segment pool expands infinitely until memory allocation fails (including heap allocations), so current param doesn't affect the "no segments error".

pasis: per ring memory footprint of a single operation is not critical info. Better to focus on the total memory footprint than single request from ring.

Total memory footprint depends on the number of sockets and their behavior (queueing). Current Total formula describes situation when every ring requests a single batch what is not how a live system look like.

Remove this section or refer to the per-socket batching param where you can describe total memory footprint based on number of sockets and queueing.

pasis: don't mention specific size of the internal structure - it can be changed.

pasis: 10-20 segments is incorrect. 10 can be correct for 2MB default send buffer and maximum TSO segments (256KB). Or, 20 is correct for default buffer size 1500 bytes and 1MB buffer size when TSO can hold maximum 32 SGEs (48KB TSO total). However, in reality, we can have significantly higher number of TCP segments (we can send 1B segments frequent on high latency network and all those segments will be in the unacked queue, or we cannot merge zc and non-zc segments, so non-zc headers break merging and creatin more segments). Also, not sure how it is today, TSO worked during tcp_output phase, not during tcp_write, what made the initial send buffer full of 1500 segments - need to check current behavior.

pasis: This is memory per request, not the tcp_seg footprint per socket. Not sure user needs this calculation since they should be interested in total memory footprint and latency/performance tradeoffs.

docs/xlio_config_reference.md
```diff
@@ -1864,8 +1886,7 @@ bytes for Ethernet/IP/TCP headers). Supports suffixes: B, KB, MB, GB.
 - Non-zero: Force specific size. Clamped to max 256KB.
   Reset to 0 if value <= Maximum Segment Size.
 
-**TCP Segmentation Offload interaction:** Controls maximum data per
-super-segment is the smaller of this buffer size and the TSO max payload size.
+**TCP Segmentation Offload interaction:** Maximum data per TSO segment is the smaller of this buffer size and the TSO max payload size.
 Without TCP Segmentation Offload, segments stay at Maximum Segment
 Size regardless of buffer size.
```
pasis: [question] Do we have any restrictions on the config subsystem level when TSO is disabled?

docs/xlio_config_reference.md
pasis: It should be increased for non-zerocopy heavy TX offloads to improve TSO utilization. Also for TLS TX offload 16KB is optimal, because the buffers are used for TLS records construction.

For zerocopy send scenarios without TLS TX offload, this should be default to reduce memory usage.

pasis (commented):
stopped at `performance.buffers.tx.buf_size`

docs/xlio_config_reference.md
pasis: AFAIR, UDP sends constant headers buf separately from the payload mem_buf_desc_t object (making it 2 SGEs send operation). Is it correct that the prefetch includes the header buffer?

pasis: looks like missed link for `adaptive_period_usec`

pasis, line 1: Consider creating something like table of contents with links to the respective params description. The idea is to quickly discover potentially interesting params without knowing their exact name.

pasis (commented):
stopped at `performance.completion_queue.interrupt_moderation.period_usec`

docs/xlio_config_reference.md
tomerdbz: I'll do that in a seperate PR - adding an x-path attribute with docs for it sounds like a great idea. fixing here tactically

tomerdbz: won't address this now - but let's discuss this

tomerdbz: no, fixing

tomerdbz: right, will change :)

tomerdbz: probably need to move it to buffer.rx group. (seperate PR)

tomerdbz: yup, fixing to make it clear now

tomerdbz: confirmed, will fix

tomerdbz: correct, my mistake - removing

tomerdbz: removed

tomerdbz: removed

docs/xlio_config_reference.md
tomerdbz: no, I'll streghten the text to emphasize this is memory waste to put large values w/o MSS. we should add something though in the future - or at least auto-correct in the code.

docs/xlio_config_reference.md
tomerdbz: In the non-inline path, everything is in one buffer(UDP header template is copied), so yes, the prefetch covers headers too.

tomerdbz: anyway removed the internl details

galnoam:
@tomerdbz , please open a task with all leftover gaps.
Please add to it also about setting structure to the file.

galnoam (approved):

PR #590, opened 2026-03-24
issue: 4043231 Fix keepalive abort to honor TCP_USER_TIMEOUT

tomerdbz:
## Description
When TCP_USER_TIMEOUT is set on a keepalive connection, the abort decision should be based on elapsed time since last received data, not on the keepalive probe count (TCP_KEEPCNT). This matches the Linux tcp_keepalive_timer() behavior where TCP_USER_TIMEOUT overrides TCP_KEEPCNT.

Previously, the keepalive abort always used keep_cnt * keep_intvl, ignoring TCP_USER_TIMEOUT entirely. The separate tcp_user_timeout_ occured() check could eventually fire but measured from the wrong reference point (first probe sent, not last data received), making the effective timeout keep_idle + user_timeout instead of just user_timeout.

Enable LWIP_TCP_KEEPALIVE so that per-socket TCP_KEEPINTVL and TCP_KEEPCNT values are honored instead of using fixed defaults.

##### What
Fix keepalive abort to honor TCP_USER_TIMEOUT and enable per-socket TCP_KEEPINTVL/TCP_KEEPCNT.

##### Why ?
issue: 4043231 - When TCP_USER_TIMEOUT is set on a keepalive connection, XLIO aborts based on
probe count (TCP_KEEPCNT) instead of elapsed time. 

This differs from Linux, where
TCP_USER_TIMEOUT overrides TCP_KEEPCNT and the connection drops when elapsed time since last
received data exceeds the user timeout. 

Additionally, TCP_KEEPINTVL and TCP_KEEPCNT socket
options were silently ignored (LWIP_TCP_KEEPALIVE was compiled out).

##### How ?
Modify the keepalive abort block in tcp_slowtmr() to branch on user_timeout_ms:
- When set: abort when elapsed-since-last-rx >= user_timeout AND at least one probe was sent
  (matches Linux tcp_keepalive_timer() logic).
- When not set: abort when probe count exhausted (existing behavior, now using per-socket values).
Enable LWIP_TCP_KEEPALIVE=1 so per-socket keep_intvl and keep_cnt fields (already initialized
from /proc/sys/net/ipv4/tcp_keepalive_* in the constructor) are used by the timer and exposed
via setsockopt/getsockopt.

## Change type
- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list

### DIFF
--- a/src/core/lwip/opt.h
+++ b/src/core/lwip/opt.h
@@ -195,15 +195,6 @@
    ---------- Socket options ----------
    ------------------------------------
 */
-/**
- * LWIP_TCP_KEEPALIVE==1: Enable TCP_KEEPIDLE, TCP_KEEPINTVL and TCP_KEEPCNT
- * options processing. Note that TCP_KEEPIDLE and TCP_KEEPINTVL have to be set
- * in seconds. (does not require sockets.c, and will affect tcp.c)
- */
-#ifndef LWIP_TCP_KEEPALIVE
-#define LWIP_TCP_KEEPALIVE 0
-#endif
-
 /* Platform endianness */
 #if !defined(__BYTE_ORDER__) || !defined(__ORDER_LITTLE_ENDIAN__) || !defined(__ORDER_BIG_ENDIAN__)
 #warning "__BYTE_ORDER__ or __ORDER_..._ENDIAN__ is not defined"
--- a/src/core/lwip/tcp.c
+++ b/src/core/lwip/tcp.c
@@ -642,6 +642,18 @@ void tcp_slowtmr(struct tcp_pcb *pcb)
                      * update without sending any data (which will force us to split the segment).
                      * tcp_zero_window_probe(pcb); */
                     tcp_keepalive(pcb);
+                    /* Anchor the TCP_USER_TIMEOUT clock for this zero-window /
+                     * persist probe: there is pending send data that cannot be
+                     * delivered because the peer's advertised window is closed.
+                     * This preserves the pre-existing behavior of aborting a
+                     * persist-stuck connection after user_timeout via
+                     * tcp_user_timeout_occured(). Pure keepalive probes (the
+                     * SOF_KEEPALIVE branch below) intentionally do NOT anchor
+                     * this - an idle connection has no unacknowledged data, so
+                     * its user_timeout is measured from last_progress_tmr. */
+                    if (pcb->ticks_since_data_sent == -1) {
+                        pcb->ticks_since_data_sent = 0;
+                    }
                 }
             } else {
                 /* Increase the retransmission timer if it is running */
@@ -714,13 +726,31 @@ void tcp_slowtmr(struct tcp_pcb *pcb)
         /* Check if KEEPALIVE should be sent */
         if ((pcb->so_options & SOF_KEEPALIVE) &&
             ((get_tcp_state(pcb) == ESTABLISHED) || (get_tcp_state(pcb) == CLOSE_WAIT))) {
-#if LWIP_TCP_KEEPALIVE
-            if ((u32_t)(tcp_ticks - pcb->tmr) >
-                tcp_ms_to_ticks(pcb->keep_idle + pcb->keep_cnt * pcb->keep_intvl))
-#else
-            if ((u32_t)(tcp_ticks - pcb->tmr) > tcp_ms_to_ticks(pcb->keep_idle + TCP_MAXIDLE))
-#endif /* LWIP_TCP_KEEPALIVE */
-            {
+            u32_t ka_elapsed = (u32_t)(tcp_ticks - pcb->tmr);
+            u8_t ka_abort = 0;
+
+            if (pcb->user_timeout_ms != 0) {
+                /* TCP_USER_TIMEOUT: abort once user_timeout has elapsed since the
+                 * connection last made forward progress (in-sequence data
+                 * received or new data acknowledged), NOT since pcb->tmr.
+                 * pcb->tmr is bumped by every received segment - including a
+                 * keepalive-probe reply (a bare dup-ACK) - which would push the
+                 * deadline out to keep_idle + user_timeout instead of just
+                 * user_timeout. last_progress_tmr is immune to those probe
+                 * replies, matching Linux tcp_keepalive_timer(), which measures
+                 * the keepalive elapsed from rcv_tstamp/lrcvtime. The keep_idle
+                 * guard ensures a probe has been attempted before we abort an
+                 * otherwise-idle connection (relevant when user_timeout is
+                 * shorter than keep_idle). */
+                u32_t ut_elapsed = (u32_t)(tcp_ticks - pcb->last_progress_tmr);
+                ka_abort = (ut_elapsed >= tcp_ms_to_ticks(pcb->user_timeout_ms)) &&
+                    (ut_elapsed >= tcp_ms_to_ticks(pcb->keep_idle));
+            } else {
+                ka_abort = (ka_elapsed >
+                            tcp_ms_to_ticks(pcb->keep_idle + pcb->keep_cnt * pcb->keep_intvl));
+            }
+
+            if (ka_abort) {
                 LWIP_DEBUGF_IP_ADDR(TCP_DEBUG,
                                     "tcp_slowtmr: KEEPALIVE timeout. Aborting connection to ",
                                     pcb->remote_ip, pcb->is_ipv6);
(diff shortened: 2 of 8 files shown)


tomerdbz:
VMA analyzer:
🔍 libvma Relevance Analyzer
────────────────────────────────────────
📁 Analyzing 2 file(s)...
⏳ Connecting to GitHub... done

  [1/2] lwip/opt.h... ✅ relevant
  [2/2] lwip/tcp.c (2 hunks)... ✅ relevant

────────────────────────────────────────
📋 Summary
────────────────────────────────────────

✅ 2 change(s) MAY BE RELEVANT to libvma:
   • src/core/lwip/opt.h
     → src/vma/lwip/opt.h
     libvma has the same feature/logic that was changed in libxlio, which is the definition of LWIP_TCP_KEEPALIVE. The change in libxlio from 0 to 1 is relevant to libvma as it enables TCP keepalive functionality.
   • src/core/lwip/tcp.c
     → src/vma/lwip/tcp.c
     libvma has the same feature/logic for TCP keepalive that was changed in libxlio, and the change appears to be a bug fix or improvement to the keepalive timer logic.

TODO - open a task for libvma

src/core/lwip/tcp.c
```diff
@@ -702,30 +702,32 @@ void tcp_slowtmr(struct tcp_pcb *pcb)
...
+            u32_t ka_elapsed = (u32_t)(tcp_ticks - pcb->tmr);
+            u8_t ka_abort = 0;
+
+            if (pcb->user_timeout_ms != 0) {
+                /* When TCP_USER_TIMEOUT is set, abort based on elapsed time
+                 * since last received data, requiring at least one probe sent.
+                 * This matches Linux tcp_keepalive_timer() behavior where
+                 * TCP_USER_TIMEOUT overrides TCP_KEEPCNT for the abort decision. */
+                u32_t user_timeout_ticks =
+                    (pcb->user_timeout_ms + slow_tmr_interval - 1U) / slow_tmr_interval;
```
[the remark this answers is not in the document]
pasis: @tomerdbz, according to [https://datatracker.ietf.org/doc/html/rfc5482#section-3.3](https://datatracker.ietf.org/doc/html/rfc5482#section-3.3), the timeout value that can be advertised via TCP header option is less than 2^31 ms (if you advertise max 15-bit value as minutes). So, we can put such limit to setsockopt(). WDYT?

src/core/lwip/opt.h
```diff
@@ -179,7 +179,7 @@
  * in seconds. (does not require sockets.c, and will affect tcp.c)
  */
 #ifndef LWIP_TCP_KEEPALIVE
-#define LWIP_TCP_KEEPALIVE 0
+#define LWIP_TCP_KEEPALIVE 1
```
pasis: you keep ability to toggle LWIP_TCP_KEEPALIVE , however, you remove build-time branching in 2 places below. So when you disable keepalive here the compilation will fail.
Either make proper compilation in all cases or remove the LWIP_TCP_KEEPALIVE entirely from the code.

tomerdbz:
libvma Relevance Analyzer
────────────────────────────────────────
📁 Analyzing 3 file(s)...
⏳ Connecting to GitHub... done

  [1/3] lwip/tcp.c (2 hunks)... ✅ relevant
  [2/3] lwip/tcp_impl.h... ✅ relevant
  [3/3] sock/sockinfo_tcp.cpp... ✅ relevant

────────────────────────────────────────
📋 Summary
────────────────────────────────────────

✅ 3 change(s) MAY BE RELEVANT to libvma:
   • src/core/lwip/tcp.c
     → src/vma/lwip/tcp.c
     libvma has the same feature/logic that was changed in libxlio, which is related to calculating user_timeout_ticks. The change in libxlio seems to be a bug fix or improvement, making the calculation more accurate. libvma has not diverged significantly in this area, making the patch applicable.
   • src/core/lwip/tcp_impl.h
     → src/vma/lwip/tcp_impl.h
     libvma has the same feature/logic that was changed in libxlio, specifically the definition of TCP timeouts. The addition of TCP_USER_TIMEOUT_MAX_MS in libxlio is relevant to libvma as it provides a maximum user timeout value as defined in RFC 5482.
   • src/core/sock/sockinfo_tcp.cpp
     → src/vma/sock/sockinfo_tcp.cpp
     libvma has the same feature/logic for handling TCP user timeouts, and the libxlio change is a relevant bug fix to prevent exceeding the maximum allowed value.

Consider applying these changes to libvma as well.

src/core/lwip/tcp.c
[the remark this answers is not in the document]
tomerdbz: I say yes, and also do the check greptile suggests

src/core/lwip/opt.h
tomerdbz: removed LWIP_TCP_KEEPALIVE

src/core/sock/sockinfo_tcp.cpp
```diff
@@ -4702,9 +4700,20 @@ int sockinfo_tcp::tcp_setsockopt(int __level, int __optname, __const void *__opt
...
-            m_pcb.user_timeout_ms = user_timeout_ms;
+            if (__optlen < sizeof(unsigned int) || !__optval) {
+                errno = EINVAL;
+                ret = -1;
+            } else {
+                unsigned int user_timeout_ms = *(unsigned int *)__optval;
+                if (user_timeout_ms > TCP_USER_TIMEOUT_MAX_MS) {
+                    si_tcp_logwarn(
+                        "TCP_USER_TIMEOUT value %u exceeds RFC 5482 maximum, clamping to %u",
+                        user_timeout_ms, TCP_USER_TIMEOUT_MAX_MS);
```
pasis: WARN level of logs will likely be printed out. This can lead to a flood in CPS scenario.
Linux kernel returns EINVAL when the value is <0 when casted to signed int. Probably we can return EINVAL instead of reducing the value?

src/core/lwip/tcp.c
```diff
@@ -702,32 +703,36 @@ void tcp_slowtmr(struct tcp_pcb *pcb)
...
-#endif /* LWIP_TCP_KEEPALIVE */
-            {
+            } else if (ka_elapsed > (pcb->keep_idle + pcb->keep_cnt_sent * pcb->keep_intvl) /
+                           slow_tmr_interval) {
                 tcp_keepalive(pcb);
-                pcb->keep_cnt_sent++;
+                if (pcb->keep_cnt_sent < UINT8_MAX) {
+                    pcb->keep_cnt_sent++;
+                }
             }
```
pasis: does it mean that when we send 255 keepalives we start sending them every timer tick regardless of the keepalive parameters? What if we configure TCP_USER_TIMEOUT to that high allowed value, can this counter reach the 8-bit limit?

tomerdbz: as discussed privately see this:
image

src/core/sock/sockinfo_tcp.cpp
tomerdbz: fixed

tomerdbz:
bot:retest

tomerdbz:
## Lab RED-GREEN evidence (issue 4043231)

- **libxlio commit**: `d739cad4` (rebased on `vNext`) - keepalive abort anchored on new `pcb->last_progress_tmr` (advanced only on real forward progress), so keepalive-probe replies no longer restart the `TCP_USER_TIMEOUT` clock.
- **Lab**: dor03 <-> dor04, offloaded ConnectX (`mlx5`), `100.123.45.0/24`, 400G, MTU 1500.
- **OFED**: OFED-internal-26.01-0.6.4.
- **Workload**: idle-keepalive repro - `SO_KEEPALIVE`, `TCP_KEEPIDLE=8s`, `TCP_KEEPINTVL=2s`, `TCP_KEEPCNT=4`, `TCP_USER_TIMEOUT=12s`; client sends 1 byte (ACKed), goes idle, peer is then blackholed (link down); measure time-to-abort from last data.
- **Result (offload-confirmed, both ends XLIO)**:
  | Build | abort time |
  |---|---|
  | Kernel (no XLIO) - correct reference | 12.1s |
  | Pre-fix (bug) | ~20.6s (= keep_idle + user_timeout) |
  | **This fix** | **12.1s / 12.1s (two trials)** = honors `TCP_USER_TIMEOUT` |
- **Local gate**: container build clean; `tests/unit_tests` 197/197 pass.
- **Risk areas**: TCP slow-timer / keepalive path; persist/zero-window re-anchors `ticks_since_data_sent` explicitly; the `user_timeout == 0` (keep_cnt-based) path is unchanged.

PR #593, opened 2026-03-25
issue: 4923648 Fix segfault: Ultra API sockets coexist w/ worker threads

tomerdbz:
## Description
When worker_threads > 0, XLIO unconditionally dispatches any socket's connect/listen/accept operations to worker threads via distribute_socket(). Ultra API sockets manage their own poll groups and rings, so this dispatch causes double-ownership: the application thread drives the socket through its poll_group while the worker thread also tries to manage it. This leads to "Buff is already a member in a list!" errors and a SIGSEGV in TCP_EVENT_ERR when the worker thread invokes a NULL callback.

Add sockinfo::should_use_threads_mode() which returns true only when worker threads are configured AND the socket is not an Ultra API socket (m_is_xlio_socket == false). Replace the three bare is_threads_mode() / worker_threads > 0 checks in sockinfo_tcp.cpp (connect, listen, accept paths) with this new method, allowing POSIX and Ultra API sockets to coexist safely when worker threads are enabled.

Add a regression test -ultra_api_worker_threads.no_distribute_on_connect that exercises xlio_socket_connect() with worker_threads=1 at debug log level and asserts that worker-thread dispatch messages are absent from the output. Extract subprocess test infrastructure from output/config.cc into common/subprocess_test.h for reuse.

##### What
Skip worker-thread dispatch for Ultra API sockets, preventing SIGSEGV when Ultra API and worker threads coexist.

##### Why ?
When worker_threads > 0, XLIO unconditionally dispatches any socket's connect/listen/accept to worker threads via distribute_socket(). Ultra API sockets manage their own poll groups and rings, so this dispatch causes double-ownership: the application thread drives the socket through its poll_group while a worker thread also tries to manage it. This leads to "Buff is already a member in a list!" errors and a SIGSEGV in TCP_EVENT_ERR when the worker thread invokes a NULL callback. See Redmine #4923648.

##### How ?
Add sockinfo::should_use_threads_mode() which returns true only when worker threads are configured and the socket is not an Ultra API socket (!m_is_xlio_socket). Replace the three bare is_threads_mode() / worker_threads > 0 checks in sockinfo_tcp.cpp (connect, listen, accept paths) with this method. POSIX sockets continue to be dispatched to worker threads as before; Ultra API sockets bypass the dispatch entirely.

A regression test (ultra_api_worker_threads.no_distribute_on_connect) launches a helper binary that calls xlio_socket_connect() with worker_threads=1 at debug log level, then asserts that worker-thread dispatch messages (connect_socket_job) are absent from the output. The test runs automatically in CI under the ultra_api* filter with real RDMA addresses and root privileges. Subprocess test infrastructure is extracted from output/config.cc into common/subprocess_test.h for reuse.

## Change type
- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connection and accept handling when worker-thread mode is enabled, making behavior more consistent across socket types.

* **Tests**
  * Added new end-to-end coverage for worker-thread connection scenarios.
  * Introduced shared test support for running helper programs and capturing their output.
  * Updated test configuration and build setup to include the new scenarios.

### DIFF
--- a/src/core/sock/sockinfo_tcp.cpp
+++ b/src/core/sock/sockinfo_tcp.cpp
@@ -2771,7 +2771,7 @@ int sockinfo_tcp::connect(const sockaddr *__to, socklen_t __tolen)
         return -1;
     }
 
-    if (safe_mce_sys().is_threads_mode()) {
+    if (should_use_threads_mode()) {
         // For Threads mode need to do partial preparation and the rest will be done by the Thread.
         connect_threads_mode();
         return -1; // Currently no blocking socket support.
@@ -3210,8 +3210,7 @@ int sockinfo_tcp::listen(int backlog)
     tcp_accepted_pcb(&m_pcb, sockinfo_tcp::accepted_pcb_cb);
 
     bool success = false;
-    // Check if XLIO threads are enforced (> 0) for entity context distribution
-    if (safe_mce_sys().worker_threads > 0) {
+    if (should_use_threads_mode()) {
         create_listen_context();
         start_sockinfo_tcp_listen_objects();
         success = wait_for_listen_rss_children_ready();
@@ -3342,8 +3341,8 @@ int sockinfo_tcp::accept_helper(struct sockaddr *__addr, socklen_t *__addrlen,
             }
         }
 
-        int tmp_ret = safe_mce_sys().worker_threads > 0 ? harvest_sockinfo_tcp_listen_objects()
-                                                        : rx_wait(poll_count, m_b_blocking);
+        int tmp_ret = should_use_threads_mode() ? harvest_sockinfo_tcp_listen_objects()
+                                                : rx_wait(poll_count, m_b_blocking);
         if (tmp_ret < 0) {
             si_tcp_logdbg("interrupted accept");
             unlock_tcp_con();
@@ -3370,7 +3369,7 @@ int sockinfo_tcp::accept_helper(struct sockaddr *__addr, socklen_t *__addrlen,
     m_ready_conn_cnt--;
     IF_STATS(m_p_socket_stats->listen_counters.n_conn_backlog--);
 
-    safe_mce_sys().worker_threads ? assert(m_syn_received.empty()) : remove_received_syn_socket(ns);
+    should_use_threads_mode() ? assert(m_syn_received.empty()) : remove_received_syn_socket(ns);
 
     unlock_tcp_con();
 
--- a/tests/gtest/common/subprocess_test.h
+++ b/tests/gtest/common/subprocess_test.h
@@ -1,0 +1,110 @@
+/*
+ * SPDX-FileCopyrightText: NVIDIA CORPORATION & AFFILIATES
+ * Copyright (c) 2022-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: GPL-2.0-only or BSD-2-Clause
+ */
+
+#ifndef TESTS_GTEST_COMMON_SUBPROCESS_TEST_H_
+#define TESTS_GTEST_COMMON_SUBPROCESS_TEST_H_
+
+#include <cstdlib>
+#include <fstream>
+#include <string>
+#include <unistd.h>
+
+#include "common/def.h"
+
+/**
+ * Base class for tests that launch helper binaries and inspect their
+ * captured output.  Provides workspace detection, command execution
+ * with output capture, and automatic cleanup.
+ */
+class subprocess_test : virtual public testing::Test {
+public:
+    void SetUp() override
+    {
+        m_workspace = std::getenv("WORKSPACE");
+        if (m_workspace) {
+            std::cout << "WORKSPACE: '" << m_workspace << "'" << std::endl;
+        } else {
+            std::cout << "WORKSPACE is not set" << std::endl;
+        }
+        // Default capture file; derived fixtures may override with a more
+        // specific name. Kept here so the base fully owns the field it cleans
+        // up in TearDown().
+        m_output_file = "/tmp/xlio_gtest_" + std::to_string(getpid()) + ".txt";
+    }
+
+    void TearDown() override
+    {
+        if (!m_output_file.empty()) {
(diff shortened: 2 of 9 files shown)


ntsemah:
bot:retest

tests/gtest/common/subprocess_test.h
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    /**
+     * Return the full path to a helper binary built alongside the gtest.
+     * Uses GTEST_BUILDDIR (set by Makefile.am) so the lookup works with
+     * out-of-tree builds where the build dir differs from the source tree.
+     */
+    std::string helper_path(const char *name) const
+    {
+        return std::string(GTEST_BUILDDIR) + "/" + name;
```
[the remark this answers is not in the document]
tomerdbz: incorrect

tomerdbz:
bot:retest

tomerdbz:
bot:retest

tests/gtest/common/subprocess_test.h
```diff
@@ -0,0 +1,105 @@
...
+/**
+ * Base class for tests that launch helper binaries and inspect their
+ * captured output.  Provides workspace detection, command execution
+ * with output capture, and automatic cleanup.
+ */
+class subprocess_test : virtual public testing::Test {
+public:
+    void SetUp() override
+    {
+        m_workspace = std::getenv("WORKSPACE");
```
orshemesh-grok, line 26: Setup() does not set m_output_file to any value, but the fixture defines it and TearDown() deletes it if it exists.
I think that this is incomplete: Either set m_output_file to some default value in Setup(), or remove it at all from this hierarchy level.

src/core/sock/sockinfo_tcp.cpp
orshemesh-grok: I agree with CodeRabbit: You should use should_use_threads_mode() here as well

orshemesh-grok (changes requested):

src/core/sock/sockinfo_tcp.cpp
tomerdbz: Done in 432a5213 - the accept cleanup now uses `should_use_threads_mode()`, matching the harvest decision at line 3344. It is a no-op for POSIX sockets (`is_threads_mode() == worker_threads > 0`) and correctly runs `remove_received_syn_socket()` for Ultra API sockets instead of asserting.

tests/gtest/common/subprocess_test.h
tomerdbz: Done in 432a5213 - `subprocess_test::SetUp()` now initializes `m_output_file` to a default (`/tmp/xlio_gtest_<pid>.txt`) so the base fully owns the field it cleans up in `TearDown()`; derived fixtures still override it with a more specific name. Also added the explicit `<unistd.h>` include it relies on.

tomerdbz:
The three open review threads here were filed against the pre-squash commit `ff44d595`. The branch was subsequently squashed and force-pushed to the current head `432a5213`, which already contains all three fixes, so the threads did not auto-resolve. Evidence (`git blame` attributes each line to `432a5213`):

1. `sockinfo_tcp.cpp` accept path now uses `should_use_threads_mode()` at both the harvest (`:3344`) and cleanup (`:3372`) sites - no raw `worker_threads`/`is_threads_mode()` check remains in that path.
2. `tests/gtest/common/subprocess_test.h`: `SetUp()` now initializes `m_output_file` (~L35), symmetric with the `TearDown()` that deletes it.
3. `tests/gtest/xlio_ultra_api/xlio_socket_worker_threads.cc` now asserts the documented dispatch marker (~L80-85), matching the docstring at L45.

@orshemesh-grok could you re-review `432a5213` and clear the change request if this looks good? Thanks.

src/core/sock/sockinfo_tcp.cpp
```diff
@@ -3342,8 +3341,8 @@ int sockinfo_tcp::accept_helper(struct sockaddr *__addr, socklen_t *__addrlen,
             }
         }
 
-        int tmp_ret = safe_mce_sys().worker_threads > 0 ? harvest_sockinfo_tcp_listen_objects()
-                                                        : rx_wait(poll_count, m_b_blocking);
+        int tmp_ret = should_use_threads_mode() ? harvest_sockinfo_tcp_listen_objects()
+                                                : rx_wait(poll_count, m_b_blocking);
```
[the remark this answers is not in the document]
tomerdbz: Already addressed in the head commit (now c944f3bf): the accept path uses should_use_threads_mode() at both the harvest (:3344) and cleanup (:3372) sites - no raw worker_threads/is_threads_mode() check remains in that path (git blame attributes both to the fix commit). This thread predates the squash/force-push. Resolving as addressed.

tests/gtest/xlio_ultra_api/xlio_socket_worker_threads.cc
```diff
@@ -0,0 +1,93 @@
...
+ *
+ *    A compiled helper binary retrieves the API via xlio_get_api(), creates a
+ *    poll group and socket, and calls xlio_socket_connect() targeting the
+ *    server address provided by the gtest runner (a real RDMA-routable address
+ *    in CI).  The config sets worker_threads=1 and log level=debug so that any
+ *    worker-thread dispatch is visible in the output.
+ *
+ *    The test asserts the ABSENCE of worker-thread dispatch messages:
+ *    - "connect_socket_job" / "New TCP socket added" from entity_context
+ *    - "already a member in a list" from vlist (symptom of the original crash)
```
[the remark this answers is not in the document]
tomerdbz: Already addressed in the head commit (now c944f3bf): the documented dispatch-marker assertion is present (~L80-85), matching the docstring. This thread predates the squash/force-push. Resolving as addressed.

orshemesh-grok (dismissed):

orshemesh-grok (approved):

PR #599, opened 2026-04-01
issue: 4043214 Honor TCP_USER_TIMEOUT over retry limits in tcp_slowtmr

tomerdbz:
## Description
When TCP_USER_TIMEOUT is set via setsockopt, the retry-count checks (TCP_SYNMAXRTX for SYN_SENT, TCP_MAXRTX for data) abort the connection before the user timeout can fire. With the default flat 3s SYN RTO and TCP_SYNMAXRTX=6, connections in SYN_SENT are killed at ~18s even when an application requested a 60s user timeout.

Per RFC 9293 Section 3.10.8, the user timeout governs conn lifetime for any state. Per Section 3.8.3(d) (MUST-21), applications must be able to set R2 for a connection. The retry-count limits are implementation heuristics for the default case and must not override an explicit user timeout.

Skip the TCP_SYNMAXRTX and TCP_MAXRTX abort checks when user_timeout_ms is non-zero, letting retransmissions continue until the user timeout fires. Cap the tcp_backoff[] index to TCP_MAXRTX to prevent an out-of-bounds read when nrtx grows beyond the array size.

Here's the filled-in template:

##### What
Fix `TCP_USER_TIMEOUT` being ignored by lwIP retry limits and cap RTO at 120s for Linux kernel parity.

##### Why ?
When `TCP_USER_TIMEOUT` is set via `setsockopt`, XLIO's lwIP stack still aborts connections after `TCP_SYNMAXRTX` (6) or `TCP_MAXRTX` (12) retransmissions — whichever comes first — ignoring the user-configured timeout entirely. This causes spurious `ETIMEDOUT` errors (e.g., at ~18s instead of the configured 60s). Per RFC 9293 Section 3.8.6.1, applications MUST be able to extend the timeout beyond default retry limits. Additionally, the RTO had no upper bound and could grow to 384+ seconds under heavy backoff, far exceeding the Linux kernel's 120s cap (RFC 6298 Section 2.5).

Relates to issue #4043214.

##### How ?
Two commits:

**1. Honor `TCP_USER_TIMEOUT` over retry limits** (`tcp.c`, `tcp_out.c`)
- Guard the `TCP_SYNMAXRTX` and `TCP_MAXRTX` abort checks in `tcp_slowtmr` with `pcb->user_timeout_ms == 0`, so when a user timeout is set, only the elapsed-time check (`tcp_user_timeout_occured`) terminates the connection.
- Harden comparisons from `==` to `>=` as a defensive measure against `nrtx` overshooting the limit.
- Cap the `tcp_backoff[]` array index using `sizeof()` to prevent out-of-bounds reads when `nrtx` grows past the array size.
- Protect `pcb->rto` (`s16_t`) from integer overflow by computing in `int` and clamping to `0x7FFF`.
- Cap `nrtx` (`u8_t`) increments at `0xFF` in both `tcp_rexmit_rto` and `tcp_rexmit` to prevent wraparound.

**2. Cap RTO at 120s** (`opt.h`, `tcp.c`)
- Define `TCP_RTO_MAX` (120000ms) in `opt.h` with `#ifndef` guard for compile-time override, placed alongside `TCP_MAXRTX`/`TCP_SYNMAXRTX`.
- Apply the cap in the backoff computation in `tcp_slowtmr`, converting to tick units at runtime (`TCP_RTO_MAX / slow_tmr_interval`). Only affects `ESTABLISHED`-state retransmissions; `SYN_SENT` uses a flat RTO without backoff and is unaffected.

## Change type
- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TCP timeout handling so connections respect user-defined timeouts more consistently.
  * Adjusted retransmission and retry behavior to better handle repeated failures and prevent premature aborts.
  * Added a cap to retransmission timing, helping avoid overly large backoff delays.
  * Prevented retry counters from overflowing during repeated retransmissions.

### DIFF
--- a/src/core/lwip/tcp.c
+++ b/src/core/lwip/tcp.c
@@ -662,8 +662,10 @@ void tcp_slowtmr(struct tcp_pcb *pcb)
                     /* Double retransmission time-out unless we are trying to
                      * connect to somebody (i.e., we are in SYN_SENT). */
                     if (get_tcp_state(pcb) != SYN_SENT) {
-                        u8_t backoff_idx =
-                            (u8_t)LWIP_MIN(pcb->nrtx, (u8_t)(ARRAY_SIZE(tcp_backoff) - 1));
+                        u8_t backoff_idx = pcb->nrtx;
+                        if (backoff_idx >= ARRAY_SIZE(tcp_backoff)) {
+                            backoff_idx = (u8_t)(ARRAY_SIZE(tcp_backoff) - 1);
+                        }
                         int calc_rto = ((pcb->sa >> 3) + pcb->sv) << tcp_backoff[backoff_idx];
                         int max_rto = TCP_RTO_MAX / (int)slow_tmr_interval;
                         calc_rto = LWIP_MIN(calc_rto, max_rto);
--- a/src/core/lwip/tcp.c
+++ b/src/core/lwip/tcp.c
@@ -50,6 +50,10 @@
 #include <sys/types.h>
 #include <unistd.h>
 
+#ifndef ARRAY_SIZE
+#define ARRAY_SIZE(arr) (sizeof(arr) / sizeof(arr[0]))
+#endif
+
 tcp_tx_pbuf_alloc_fn external_tcp_tx_pbuf_alloc;
 tcp_tx_pbuf_free_fn external_tcp_tx_pbuf_free;
 tcp_rx_pbuf_free_fn external_tcp_rx_pbuf_free;
@@ -580,8 +584,7 @@ static inline bool tcp_user_timeout_occured(struct tcp_pcb *pcb)
     u32_t user_timeout_ticks = (pcb->user_timeout_ms + slow_tmr_interval - 1U) / slow_tmr_interval;
 
     return pcb->user_timeout_ms != 0 && pcb->ticks_since_data_sent > 0 &&
-        (u32_t)pcb->ticks_since_data_sent > user_timeout_ticks &&
-        (get_tcp_state(pcb) == ESTABLISHED || get_tcp_state(pcb) == SYN_SENT);
+        (u32_t)pcb->ticks_since_data_sent > user_timeout_ticks;
 }
 
 /**
@@ -620,11 +623,12 @@ void tcp_slowtmr(struct tcp_pcb *pcb)
             err = ERR_TIMEOUT;
             pcb_reset += (pcb->so_options & SOF_KEEPALIVE);
             LWIP_DEBUGF(TCP_DEBUG, ("tcp_slowtmr: user timeout occurred\n"));
-        } else if (get_tcp_state(pcb) == SYN_SENT && pcb->nrtx == TCP_SYNMAXRTX) {
+        } else if (get_tcp_state(pcb) == SYN_SENT && pcb->nrtx >= TCP_SYNMAXRTX &&
+                   pcb->user_timeout_ms == 0) {
             ++pcb_remove;
             err = ERR_TIMEOUT;
             LWIP_DEBUGF(TCP_DEBUG, ("tcp_slowtmr: max SYN retries reached\n"));
-        } else if (pcb->nrtx == TCP_MAXRTX) {
+        } else if (pcb->nrtx >= TCP_MAXRTX && pcb->user_timeout_ms == 0) {
             ++pcb_remove;
             err = ERR_ABRT;
             LWIP_DEBUGF(TCP_DEBUG, ("tcp_slowtmr: max DATA retries reached\n"));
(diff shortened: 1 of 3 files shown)


tomerdbz:
libvma Relevance Analyzer
────────────────────────────────────────
📁 Analyzing 3 file(s)...
⏳ Connecting to GitHub... done

  [1/3] lwip/opt.h... ✅ relevant
  [2/3] lwip/tcp.c (2 hunks)... ✅ relevant
  [3/3] lwip/tcp_out.c (2 hunks)... ✅ relevant

────────────────────────────────────────
📋 Summary
────────────────────────────────────────

✅ 3 change(s) MAY BE RELEVANT to libvma:
   • src/core/lwip/opt.h
     → src/vma/lwip/opt.h
     libvma has TCP options and the change in libxlio is related to the maximum retransmission timeout, which is a TCP setting. The change seems to be a bug fix or improvement, making it relevant to libvma.
   • src/core/lwip/tcp.c
     → src/vma/lwip/tcp.c
     libvma has the same feature/logic that was changed in libxlio, which is the handling of TCP state and retransmission counters. The change in libxlio seems to be a bug fix or improvement, adding a check for user_timeout_ms, which is also present in libvma.
   • src/core/lwip/tcp_out.c
     → src/vma/lwip/tcp_out.c
     libvma has the same feature/logic that was changed in libxlio, which is the increment of the nrtx variable. The change in libxlio is a bug fix to prevent overflow, and this bug fix is also relevant to libvma.

Consider applying these changes to libvma as well.

src/core/lwip/tcp.c
```diff
@@ -620,11 +620,12 @@ void tcp_slowtmr(struct tcp_pcb *pcb)
...
             pcb_reset += (pcb->so_options & SOF_KEEPALIVE);
             LWIP_DEBUGF(TCP_DEBUG, ("tcp_slowtmr: user timeout occurred\n"));
-        } else if (get_tcp_state(pcb) == SYN_SENT && pcb->nrtx == TCP_SYNMAXRTX) {
+        } else if (get_tcp_state(pcb) == SYN_SENT && pcb->nrtx >= TCP_SYNMAXRTX &&
+                   pcb->user_timeout_ms == 0) {
             ++pcb_remove;
             err = ERR_TIMEOUT;
             LWIP_DEBUGF(TCP_DEBUG, ("tcp_slowtmr: max SYN retries reached\n"));
-        } else if (pcb->nrtx == TCP_MAXRTX) {
+        } else if (pcb->nrtx >= TCP_MAXRTX && pcb->user_timeout_ms == 0) {
```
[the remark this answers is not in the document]
tomerdbz: Great!

src/core/lwip/tcp.c
```diff
@@ -658,7 +658,12 @@ void tcp_slowtmr(struct tcp_pcb *pcb)
                     /* Double retransmission time-out unless we are trying to
                      * connect to somebody (i.e., we are in SYN_SENT). */
                     if (get_tcp_state(pcb) != SYN_SENT) {
-                        pcb->rto = ((pcb->sa >> 3) + pcb->sv) << tcp_backoff[pcb->nrtx];
+                        u8_t backoff_idx =
+                            (u8_t)LWIP_MIN(pcb->nrtx, (u8_t)(sizeof(tcp_backoff) - 1));
```
pasis: Use ARRAY_SIZE() instead of sizeof(). If the type of the array is changed, it can introduce OOB silently.

tomerdbz: great, fixed another nearby occurence as well.
note though utils.h where the ARRAY_SIZE macro resides is mostly c++, so I'd just re-define the macro. let me know if you agree with this appraoch

src/core/lwip/tcp.c
```diff
@@ -50,6 +50,10 @@
 #include <sys/types.h>
 #include <unistd.h>
 
+#ifndef ARRAY_SIZE
+#define ARRAY_SIZE(arr) (sizeof(arr) / sizeof(arr[0]))
+#endif
```
pasis: put it to some header instead of C file to reuse it across lwip code.

tomerdbz:
bot:retest

src/core/lwip/tcp.c
tomerdbz: Done - `tcp_persist_backoff` now uses `ARRAY_SIZE()` instead of `sizeof()`, and the new `tcp_backoff[]` index added in this PR uses it as well.

src/core/lwip/tcp.c
tomerdbz: Done - moved `ARRAY_SIZE` to `src/core/lwip/opt.h` (included by all lwip sources) and removed the local definition from `tcp.c`, so it is shared across the lwip code as suggested.

tomerdbz:
## Testing / how to reproduce (not covered by automated CI)

This path needs a lab pair with a **blackholed peer**, so it isn't exercised by CI. To verify:

1. On two XLIO-offloaded hosts, create a socket and set `TCP_USER_TIMEOUT` (e.g. 60s); leave `TCP_SYNMAXRTX` / `TCP_MAXRTX` at defaults.
2. **SYN_SENT case**: `connect()` to an address that never answers the SYN (firewalled/blackholed peer).
   **Established-data case**: connect, send data, then blackhole the peer (drop its traffic / down the link).
3. Measure time-to-error (`recv()` returning / `SO_ERROR`).

**Expected:**
- Without this change: the connection aborts at the retry cap (~`TCP_SYNMAXRTX` x RTO ≈ 18s for SYN with the 3s base RTO), **ignoring** `TCP_USER_TIMEOUT`.
- With this change: retransmits continue until `TCP_USER_TIMEOUT` (~60s), per RFC 9293 §3.10.8 / Linux parity.

pasis (approved):

PR #601, opened 2026-04-06
issue: 4962651 Fix "use xlio" keyword rejected by rules parser

tomerdbz:
## Description
The pre-generated flex scanner DFA tables were never regenerated after the VMA-to-XLIO rename. The rename was a text find-and-replace that updated action code (return XLIO) but left the numeric DFA transition tables encoding the old "vma" keyword. Character 'x' remained in equivalence class 1 (catch-all), making "use xlio" impossible to tokenize -- the lexer returned raw 'x' causing a parse error.

Regenerate config_scanner.c and config_parser.c/h from the .l/.y sources using flex 2.6.4 and bison 3.8.2. Fix const mismatch in config_parser.y for __xlio_parse_config_line signature.

##### What
Fix "use xlio" keyword rejected by acceleration rules parser.

##### Why ?
use xlio tcp_client <addr>:<port>:*:* in acceleration rules (both legacy .conf and new JSON config) produces syntax error, unexpected unrecognized-token, while use os works fine. 

Root cause: c and h files were never regenerated to comply to config_parser.y

##### How ?
architecture, approach, etc._

## Change type
- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list

### DIFF
--- a/src/core/util/config_parser.y
+++ b/src/core/util/config_parser.y
@@ -51,7 +51,7 @@ int __xlio_config_empty(void)
 /* define the address by 4 integers */
 static void __xlio_set_ipv4_addr(short a0, short a1, short a2, short a3)
 {
-	char buf[16];
+	char buf[32];
 	struct in_addr *p_ipv4 = NULL;
   
 	p_ipv4 = &(__xlio_address_port_rule->ipv4);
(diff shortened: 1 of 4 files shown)


BasharRadya (approved):
CI build fails,

no new code on non-auto generated files, just regenerate auto-generated files.

BasharRadya (changes requested):
build fails

BasharRadya (approved):
Change only in the Auto generated files (generated by our src fiels .y etc..),
no impactful change in the src files.

PR #610, opened 2026-04-16
issue 4938837: [XLIO][Ultra API] - Call error callback only after accept callback

orshemesh-grok:
## Description
Make sure Ultra-API callbacks for Events are made only for sockets which the application is aware of.
An application is aware of sockets it created (i.e. connect(), listen()), and sockets that it learned about using the accept callback.

##### Why ?
The contract between XLIO Ultra API and an application is that the application gets callbacks only of sockets it is aware of.
This way, the app can be sure that when it sets user-data for sockets it creates and sockets it accepts, all callbacks will have a valid user-data value.

##### How ?
Add sockinfo_tcp::m_xlio_app_callbacks_allowed, which is set to true for sockets the application is aware of, set is accordingly and use it to gate callbacks.

## Change type
- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list
- [X] Documentation has been updated (if necessary)

## Release Notes

* **Bug Fixes**
  * Improved socket initialization and attachment behavior to more reliably control callback execution timing during socket lifecycle operations, particularly for socket creation, acceptance, and group attachment scenarios.

### DIFF
--- a/src/core/sock/sockinfo_tcp.cpp
+++ b/src/core/sock/sockinfo_tcp.cpp
@@ -392,7 +392,12 @@ void sockinfo_tcp::rx_add_ring_cb(ring *p_ring)
     sockinfo::rx_add_ring_cb(p_ring);
 }
 
-void sockinfo_tcp::set_xlio_socket(const struct xlio_socket_attr *attr)
+bool sockinfo_tcp::xlio_app_callbacks_allowed() const
+{
+    return is_xlio_socket() && m_p_group && m_xlio_app_callbacks_allowed;
+}
+
+void sockinfo_tcp::set_xlio_socket(const struct xlio_socket_attr *attr, bool allow_app_callbacks)
 {
     if (is_shadow_socket_present() && m_rx_epfd != -1) {
         // XLIO Socket API doesn't use per socket epfd
@@ -422,6 +427,7 @@ void sockinfo_tcp::set_xlio_socket(const struct xlio_socket_attr *attr)
     tcp_acked(&m_pcb, nullptr);
     set_blocking(false);
     m_is_xlio_socket = true;
+    m_xlio_app_callbacks_allowed = allow_app_callbacks;
 }
 
 void sockinfo_tcp::set_entity_context(entity_context *ctx)
@@ -612,7 +618,7 @@ int sockinfo_tcp::attach_xlio_group(poll_group *group)
 
     // TODO reinitialize lwip callbacks
 
-    set_xlio_socket(&attr);
+    set_xlio_socket(&attr, true /* Allow app callbacks */);
     group->add_socket(this);
 
     std::lock_guard<decltype(m_tcp_con_lock)> lock(m_tcp_con_lock);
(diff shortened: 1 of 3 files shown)


src/core/sock/sockinfo_tcp.cpp
```diff
@@ -612,7 +623,9 @@ int sockinfo_tcp::attach_xlio_group(poll_group *group)
 
     // TODO reinitialize lwip callbacks
 
-    set_xlio_socket(&attr);
+    // Make sure we are not raising the flag, we assume it is already set */
+    assert(m_xlio_app_callbacks_allowed);
+    set_xlio_socket(&attr, true /* Allow app callbacks */);
```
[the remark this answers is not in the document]
orshemesh-grok: Fixed

src/core/sock/sockinfo_tcp.cpp
```diff
@@ -685,6 +693,10 @@ err_t sockinfo_tcp::rx_lwip_cb_xlio_socket(void *arg, struct tcp_pcb *pcb, struc
     tcp_recved(pcb, p->tot_len, true);
 
     if (conn->m_p_group->m_socket_rx_cb) {
+        if (!conn->xlio_app_callbacks_allowed()) {
+            pbuf_free(p);
```
BasharRadya: should we even reach such thing?
this breaks the TCP stream integrity
@pasis @orshemesh-grok

src/core/sock/sockinfo_tcp.cpp
```diff
@@ -5790,7 +5805,8 @@ void sockinfo_tcp::tcp_express_zc_callback(mem_buf_desc_t *p_desc)
     sockinfo_tcp *si = reinterpret_cast<sockinfo_tcp *>(p_desc->tx.zc.ctx);
     const uintptr_t opaque_op = reinterpret_cast<uintptr_t>(p_desc->lwip_pbuf.desc.opaque);
 
-    if (opaque_op && si->m_p_group && si->m_p_group->m_socket_comp_cb) {
+    if (opaque_op && si->m_p_group && si->m_p_group->m_socket_comp_cb &&
+        si->xlio_app_callbacks_allowed()) {
```
BasharRadya: i don't see we need this.

we get a completion only if user sent data via the socket, which indicates that the user knows about the socket, so its trivial that the callbacks are allowed.

orshemesh-grok: I agree, removed it.

src/core/sock/sockinfo_tcp.h
```diff
@@ -391,7 +391,8 @@ class sockinfo_tcp : public sockinfo {
     int tcp_tx_express_inline(const struct iovec *iov, unsigned iov_len, unsigned flags);
     void flush();
     void make_dirty();
-    void set_xlio_socket(const struct xlio_socket_attr *attr);
+    // Sets XLIO socket attributes and status of application callbacks gating
+    void set_xlio_socket(const struct xlio_socket_attr *attr, bool allow_app_callbacks);
```
BasharRadya, line 395: lets change to bool outgoing conneciton

src/core/sock/sockinfo_tcp.cpp
orshemesh-grok: You are right, and anyway we should not have RX packets before establishment and accept notification.
There is also the performance implication.
I removed the check.

PR #615, opened 2026-04-23
issue: 4934380 Improve config reference accuracy

pasis:
## Description
Rewrite several parameter descriptions in docs/xlio_config_reference.md to match current implementation behavior, correcting outdated text and tightening tuning guidance.

##### What
Improve accuracy in docs/xlio_config_reference.md

##### Why ?
Documentation accuracy improvement.

## Change type
- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [x] Documentation content changes
- [ ] Tests
- [ ] Other

## Check list
- [x] Documentation has been updated (if necessary)

## Documentation

* Revised configuration parameter documentation with clarified behavioral descriptions, updated performance and memory guidance, and enhanced monitoring references for completion queues, polling mechanisms, ring allocation, and threading settings.


docs/xlio_config_reference.md
```diff
@@ -3895,32 +3960,38 @@ lock overhead. Misuse symptoms: hung connections
...
 
 **Sizing:** For most workloads the default (10) is
-sufficient. Increase to 32 only for high-connection-count
-servers where internal-thread CPU is measurable (NGINX
-and NGINX_DPU profiles use 32). Pair with a proportional
-[`network.protocols.tcp.timer_msec`](#networkprotocolstcptimer_msec) (for example, 32 with [`network.protocols.tcp.timer_msec`](#networkprotocolstcptimer_msec) 256).
+sufficient. Increase for high-connection-count
+servers where internal-thread CPU is measurable
+
+**Auto-set to 32 by** nginx and ngind_dpu profiles.
```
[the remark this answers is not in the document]
pasis: fixed

docs/xlio_config_reference.md
```diff
@@ -2608,30 +2631,44 @@ rate. At socket close, XLIO logs "Rx byte : max X / dropped Y (Z%)".
...
+
+- *Too low:* Elevated latency on idle-to-active transitions;
+  dominated by interrupt + wakeup cost.
+- *Too high / -1:* High CPU utilization while idle, starving
+  co-located threads.
 
 **Auto-modified by:** ultra_latency (-1), latency (-1).
 
+**Monitoring:** In xlio_stats check "Rx poll: hit/miss" -- a high
+miss rate means the budget is wasted spinning before each sleep
```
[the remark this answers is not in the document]
pasis: fixed

docs/xlio_config_reference.md
```diff
@@ -2392,33 +2392,47 @@ using very low values.
...
+  traffic and throughput workloads. Risk: visible software drops
+  appear immediately under buffer pressure instead of being
+  deferred.
+- *false*: No drops until debt reaches the full RQ depth, so
+  latency and jitter are more predictable under transient pressure.
+  Risk: the RQ may temporarily run with fewer posted WQEs, so
+  sustained buffer starvation can lead to hardware drops. Mitigate
+  with larger [`spare_buffers`](#performanceringsrxspare_buffers),
+  [`ring_elements_count`](#performanceringsrxring_elements_count),
+  or [`core.resources.memory_limit`](#coreresourcesmemory_limit).
```
tomerdbz: as a customer, which should I pick between them and why? (I think memory_limit first right? then ring_elements_count and only then spare_buffers?)

docs/xlio_config_reference.md
```diff
@@ -2466,23 +2479,28 @@ consider increasing.
...
+TCP stalls. If the application already drained the Completion Queue,
+the thread returns immediately (no redundant work).
 
-**Sizing:** Match to the longest gap between the application's
-socket API calls. Tight event loops or busy-poll should disable (0).
+This feature guarantees TCP progress and avoids RQs overflow for
+non-networking-oriented applications. However, it doesn't solve the problem
+of maximizing networking throughput for such applications.
+
+**Sizing:** Should be shorter than 200ms (minimum TCP RTO).
```
tomerdbz: is 200ms really the min TCP RTO (in RFC?) or just linux min RTO?

docs/xlio_config_reference.md
```diff
@@ -2575,10 +2598,10 @@ TCP sockets (TCP uses flow control via the receive window instead).
...
 **Behavior:** UDP packets queue until the application calls recv().
 When queued bytes reach the limit, new packets are silently dropped.
-effective_limit = max(SO_RCVBUF x 2, `override_rcvbuf_limit`)
-(Linux doubles SO_RCVBUF internally; XLIO applies the same convention.)
-Value of 0 disables the minimum; the application's SO_RCVBUF is used
-directly.
+effective_limit = max(SO_RCVBUF, `override_rcvbuf_limit`).
+Value of 0 disables the minimum.
+
+Note, setsockopt(SO_RCVBUF) doubles the user-supplied value internally.
```
tomerdbz, line 2604: nit/style: I feel the note is more confusing

docs/xlio_config_reference.md
```diff
@@ -2608,30 +2631,44 @@ rate. At socket close, XLIO logs "Rx byte : max X / dropped Y (Z%)".
...
+- **0:** Promoted to 1 (at least one poll before sleeping).
 
-**Symptoms:** Too high: receive thread at 100% CPU during idle.
+**Sizing:** Higher values trade CPU for lower wakeup latency;
+lower values free CPU at the cost of more interrupt/sleep
+transitions.
 
-**Too low:** elevated latency; in xlio_stats full view (mode 3),
-"Rx poll: X / Y (Z%) [miss/hit]" shows >50% miss rate.
+**Cause-effect:**
```
tomerdbz, line 2655: I'm not against - Cause-effect, but it just shows we need structure for the next iteration (Schema shouldn't have a description, but "sizing",  "symptoms", etc.

docs/xlio_config_reference.md
```diff
@@ -2878,19 +2906,17 @@ so the socket wakes when packets arrive on a shared ring.
...
+- *true*: Registered before sleep, removed on wakeup. Adds two extra
+  epoll_ctl() syscalls per blocking socket call. Scales at any connection
+  count but adds overhead.
 
 Enable when many sockets share few rings (per_interface or per_thread
-allocation) and connections exceed ~10,000. Symptoms: high kernel CPU
-with many idle connections, latency spikes scaling with connection
-count. For fewer than 10,000 connections, per_socket ring allocation,
-or primarily non-blocking I/O, the default (false) is sufficient.
+allocation). Symptoms: high kernel (SoftIRQ) CPU with many idle
```
tomerdbz: \n and ** on Symptoms?

docs/xlio_config_reference.md
```diff
@@ -3020,13 +3047,10 @@ multiple threads share CPU cores.
 **Protocol behavior:**
 
 - *UDP:* Yields every N iterations (at N-1, 2N-1, ...).
-- *TCP:* No yield for the first N iterations; yields on
-  every iteration thereafter.
+- *TCP:* Currently has effect only in the Threads Mode.
```
tomerdbz, line 3051: we ok on omitting: "No yield for the first N iterations; yields on
  every iteration thereafter."?

docs/xlio_config_reference.md
```diff
@@ -3453,15 +3483,27 @@ on virtual machines and when the NIC has no device memory.
...
+which equals to size_limit / system_page_size.
+Exceeding the limits auto-disables the device memory for
+the extra rings with a non-critical warning.
+
+For example, for the regular 4KB system page size and
+dual-port NIC, the maximum number of rings with enabled
+device memory is 32 per port. It can be reached with
+`max_on_device_memory` = 4KB.
+
+**Auto-set to 16384** by the latency and ultra_latency profiles.
```
tomerdbz: great, it's also auto-set by nvme_bf3 profile to 1024

docs/xlio_config_reference.md
```diff
@@ -2392,33 +2392,47 @@ using very low values.
...
+what happens when buffers are exhausted at that point:
+
+- *true* (default): Drop the newest incoming packet and recycle
+  its buffer back into the RQ so the queue stays at full depth.
+  Each drop is counted as "SW RX Packets dropped:" in xlio_stats.
+- *false*: Deliver the packet to the application and let the RQ
+  run below full depth until buffers are freed. A forced drop
+  only occurs if the shortage persists long enough that the RQ
+  would otherwise stall entirely (the consumed-slot count reaches
+  [`ring_elements_count`](#performanceringsrxring_elements_count)).
```
tomerdbz: do we want to say it's rounded up to the the power of 2?

docs/xlio_config_reference.md
```diff
@@ -2608,30 +2631,44 @@ rate. At socket close, XLIO logs "Rx byte : max X / dropped Y (Z%)".
...
-switching to interrupt-driven sleep. Despite the name, this is
-a loop count, not microseconds.
+Busy-poll budget for the blocking RX path. Despite the name,
+this is a loop count, not microseconds — the actual wall-time
+spent polling depends on CPU speed and Completion Queue state.
 
-On blocking recv()/read(), XLIO loops up to this count checking
-for completions (sub-microsecond return on hit). When exhausted,
-arms interrupts and sleeps (added wakeup latency). Non-blocking sockets always perform exactly 1 poll.
+When a blocking recv()/read()/accept()/connect() finds no data
```
tomerdbz, line 2638: is this relevant for connect()?

docs/xlio_config_reference.md
```diff
@@ -3790,11 +3854,11 @@ sufficient.
 CPU core(s) on which the XLIO internal thread runs.
 
 **Format:** same as taskset — hex bitmask (0x0F) or
-comma-delimited list with ranges (0,2,4-7). The
-internal thread processes TCP retransmission/keepalive
-timers, ARP resolution, fragment reassembly, socket
```
tomerdbz, line 3793: it doesn';t do IP fragment reassembly for UDP?

docs/xlio_config_reference.md
```diff
@@ -3930,11 +4001,7 @@ and NGINX_DPU profiles use 32). Pair with a proportional
 >
 > **Maps to:** `XLIO_MULTILOCK`
 
-Selects spinlock or mutex for XLIO's ring, socket,
-TCP connection, and device locks. On the hot path
```
tomerdbz, line 3932: why did you remove the "on the hot path..."?

src/core/config/descriptor_providers/xlio_config_schema.json
```diff
@@ -753,59 +753,59 @@
...
...ription": "Maps to XLIO_RING_MIGRATION_RATIO_TX environment variable.\nSend operations between checks for whether a socket's\ntransmit ring should migrate to the current thread.\nOnly effective with per_thread, per_cpuid, or per_core\n[`allocation_logic`](#performanceringstxallocation_logic).\nAuto-disabled when TCP Segmentation Offload\n(`hardware_features.tcp.tso.enable`)\nis enabled.\n\n**Behavior:** Every (value) sends, XLIO compares the\nsending thread to the ring owner. On mismatch, the new\nthread must stay stable for 20 more sends before\nmigration. Migration acquires both ring locks, may\nreallocate Queue Pairs, and invalidates cached state.\n\n**Tradeoffs:**\n\n- *-1 (default):* Disabled. Socket stays on the\n  creating thread's ring permanently. Zero overhead.\n- *Positive:* Adapts ring to the active thread. Lower\n  values detect sooner (more per-send overhead); higher\n  values check less often (slower adaptation).\n\n**Sizing:** Most applications pin sockets to their\ncreating thread — use -1. Enable (50-500) for\nwork-stealing or hand-off patterns where a socket\npermanently moves threads.\n\n**Symptom of wrong setting:** cross-CPU cache misses\nand lock contention."
```
tomerdbz: there is a gotcha here that was missed during the conversion
the schema doesn't have markdown links - they are generated by the script.

the schema has `symbol` notation - and the script makes it a link

src/core/config/descriptor_providers/xlio_config_schema.json
```diff
@@ -753,59 +753,59 @@
...
...iptors (Work Request\nElements) per transmit ring. Determines how many packets\ncan be in-flight (posted to the NIC, not yet completed).\n\n**Behavior:** When the Send Queue fills before completions\nfree space, behavior depends on the protocol:\n\n- *UDP:* Packet is dropped.\n- *TCP:* Packet is not sent and retry happens on either send() call,\n  incoming TCP segment or RTO. Without subsequent send() calls,\n  recovery can take RTT or RTO time in corner cases.\n\n**Tradeoffs:**\n\n- *Low (256):* Less memory, faster queue drain, more\n  predictable latency. But higher risk of Send Queue full\n  when sends outpace completions.\n- *High (32768, default):* Absorbs send spikes, more\n  packets in flight, higher sustained throughput. But more\n  memory and longer completion processing cycles.\n\n**Sizing:** For throughput workloads the default (32768) is\nsufficient. For latency-sensitive workloads (trading,\nreal-time), start with 256 (LATENCY and ULTRA_LATENCY\nprofiles) and increase only if \"TX Dropped Send Reqs:\" in\nxlio_stats -v 3 (full view) is non-zero.\n\nAutomatically capped to the NIC's maximum Queue Pair work\nrequests, and raised to at least\n`completion_batch_size` × 2."
```
tomerdbz, line 776: why did you remove the extra \n in \n\n**Send Queue depth:**?

src/core/config/descriptor_providers/xlio_config_schema.json
```diff
@@ -923,7 +923,7 @@
...
...ods are configured separately (TCP\nretransmission/keepalive/delayed-ACK cadence is set via\n`network.protocols.tcp.timer_msec`).\n\nSubsystems whose periodic/one-shot timers are paced by this resolution:\n\n- TCP timer wheel and LwIP tick (see\n  `network.protocols.tcp.timer_msec`)\n- Neighbour discovery / ARP resolution and retries\n- Netlink route/link periodic refresh\n- Completion-queue periodic drain\n  (`performance.completion_queue.periodic_drain_msec`)\n- CQ adaptive interrupt-moderation updates\n  (`performance.completion_queue.interrupt_moderation.adaptive_change_frequency_msec`)\n\n**Tradeoffs:**\n\n- *Low (10, default):* Fine-grained internal-thread tick.\n  The TCP timer wheel advances in small increments. But more `epoll_wait`\n  wakeups and higher baseline CPU on the internal thread.\n- *High (32+):* Fewer wakeups and lower internal-thread\n  CPU. But every timer-driven subsystem becomes coarser.\n  The TCP timer wheel fires more per-socket TCP timers\n  at once.\n\n**Sizing:** For most workloads the default (10) is\nsufficient. Increase for high-connection-count\nservers where internal-thread CPU is measurable\n\n**Auto-set to 32 by** nginx and nginx_dpu profiles."
```
tomerdbz: add dot.
"CPU is measurable" should be "CPU is measurable."

src/core/config/descriptor_providers/xlio_config_schema.json
```diff
@@ -1017,29 +1017,29 @@
...
...before arming interrupts and sleeping.\n\n**Mechanism:** Polls the Completion Queue for up to this many\nmicroseconds (capped by the application timeout). On expiry\nwithout data, arms Completion Queue interrupts and sleeps.\n\n- **-1:** Infinite polling; never sleeps; 100% CPU. Auto-disables\n  Completion Queue interrupt moderation.\n- **0:** Single poll then sleep (interrupt-driven). Set by\n  the nginx profile (`profiles.spec`).\n\n**Sizing:** Match to expected packet inter-arrival time.\nMonitor \"Polls [miss/hit]\" in xlio_stats: a high miss\nrate means the budget exceeds traffic frequency, wasting\nCPU spinning before every sleep transition.\n\n**Cause-effect:**\n\n- *Too low:* Frequent misses cause interrupt + context-switch\n  latency on each wakeup.\n- *Too high / -1:* Thread spins idle, starving co-located\n  threads on the same core.\n\nFor most workloads the default (100000 = 100 milliseconds)\nis sufficient. Use -1 only with dedicated CPU cores.\n\n**Auto-modified by:** ultra_latency (-1), latency (-1), nginx (0),\nnginx_dpu(0), threads mode (-1).\n\n**Related:** `blocking_rx_poll_usec` (same for recv()/read()),\n`poll_os_ratio` (OS polling within the busy-poll loop)."
```
tomerdbz: nginx_dpu(0) to nginx_dpu (0)

src/core/config/descriptor_providers/xlio_config_schema.json
```diff
@@ -1017,29 +1017,29 @@
...
...nControls how often the OS gets first-priority polling at the\nstart of select()/poll()/epoll_wait(), before checking\noffloaded sockets.\n\n**Mechanism:** A counter resets to this value after each\nOS-priority poll and decrements every iomux call. At zero,\nthe OS is polled first; otherwise `poll_os_ratio` governs\nthe Completion Queue-to-OS ratio within the polling loop.\n\n**Effective frequency:** one OS-priority poll every N calls.\n\n**Sizing:**\n\n- **0 or 1:** OS gets priority every call (most overhead).\n- **Higher values:** Fewer OS-priority polls per sequence.\n\n**Cause-effect:**\n\n- *Too low:* Syscall on each iomux entry; offloaded packets\n  wait while OS is checked first.\n- *Too high:* Non-offloaded file descriptors (timers,\n  signals, pipes) wait longer for priority checks.\n\nFor most mixed workloads the default (4, meaning OS\npriority every fourth call) is sufficient. Increase for\noffloaded-dominant workloads; decrease if non-offloaded\nresponsiveness is critical.\n\n**Auto-modified by:** ultra_latency (0), nginx (1000), nginx_dpu (1000)\n\n**Related:** `poll_os_ratio` (Completion Queue-to-OS ratio\nwithin polling loops), `poll_usec` (polling phase duration)."
```
tomerdbz: , nginx_dpu (1000) should be , nginx_dpu (1000).

docs/xlio_config_reference.md
```diff
@@ -2788,9 +2819,6 @@ before the limit. Default (16) suits most workloads.
 **Set to 128** by the nginx profile
 ([`profiles.spec`](#profilesspec)=nginx).
 
-**Related:** [`blocking_rx_poll_usec`](#performancepollingblocking_rx_poll_usec) (recv() poll duration),
```
tomerdbz, line 2789: why did you remove this?

tomerdbz (changes requested):
small stuff and ready to merge! :)

docs/xlio_config_reference.md
pasis: the ring_elements_count? I'd prefer to have those specifics in that parameter description. That's why we have links - to jump quickly if we want to understand more. We can also change constrains and auto adjustments for ring_elements_count and then information will be outdated here. Would be nice to have single point of knowledge, so it will be easier to keep documentation accurate and consistent.

docs/xlio_config_reference.md
pasis: I would say the order you mentioned is right and sorted by simplicity of understanding of the impact. Will update.

docs/xlio_config_reference.md
pasis: yes, this is workaround specifically for the threads mode to support some blocking operations while we enlist it as a gap. And implementation looks buggy, most likely it's supposed to yield every N iterations. So it's better to fix it in the code or rework as part of blocking IO task.

docs/xlio_config_reference.md
pasis: yes, we need to define structure and tags explicitly and make the document consistent. There is inconsistency across tag names (for example, Auto-set to by, Auto-modified by, etc). With the strict structure and tags it will become searchable and better suitable for AI instructions.

docs/xlio_config_reference.md
pasis: you're right, this is linux min rto

pasis: updated to the minimum linux rto

docs/xlio_config_reference.md
pasis: yes, blocking connect() needs to wait for 3-way handshake. "no data ready" is probably confusing for accept() and connect(). Need to update description

docs/xlio_config_reference.md
```diff
@@ -2706,7 +2741,8 @@ CPU spinning before every sleep transition.
 For most workloads the default (100000 = 100 milliseconds)
 is sufficient. Use -1 only with dedicated CPU cores.
 
-**Auto-set to -1** in threads mode.
+**Auto-modified by:** ultra_latency (-1), latency (-1), nginx (0),
+nginx_dpu(0), threads mode (-1).
```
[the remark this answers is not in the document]
pasis: fixed

docs/xlio_config_reference.md
```diff
@@ -2466,23 +2479,28 @@ consider increasing.
...
+TCP requires execution context for ACKs and TCP progress. If the
+application has idle phases between socket calls, this timer prevents
+TCP stalls. If the application already drained the Completion Queue,
+the thread returns immediately (no redundant work).
 
-**Sizing:** Match to the longest gap between the application's
-socket API calls. Tight event loops or busy-poll should disable (0).
+This feature guarantees TCP progress and avoids RQs overflow for
+non-networking-oriented applications. However, it doesn't solve the problem
+of maximizing networking throughput for such applications.
```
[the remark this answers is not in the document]
pasis: Added "Pair with:" description regarding periodic_drain_max_cqes.

docs/xlio_config_reference.md
pasis: done

docs/xlio_config_reference.md
pasis: nice catch, added. Btw, looks like it's pointless to set device memory lower than system page size, maybe need to revisit the profile.

docs/xlio_config_reference.md
pasis: max_rx_poll_batch is general limit for any type of polling (busy polling, interrupt, single poll before interrupt, etc). It's task is to avoid bursts of RX handling, smoothen latency, avoid queues overflows and reduce inflight resources across sockets.

On the other hand, blocking_rx_poll_usec is about CPU usage and latency and it defines specific type of polling. To me they're not strictly related, just work in the same stacktrace.

docs/xlio_config_reference.md
pasis: XLIO doesn't support IP fragmentation. It can be achieved for UDP via kernel receive on XLIO offloaded socket. Even if we created some special steering for IP fragments (with no L4 headers), the reassembly would be done synchronously on packets receive and the time (fired by internal thread) only discards expired incomplete fragments.

docs/xlio_config_reference.md
pasis: because this is not true for all the hot/data paths. For example, sockets are locked unconditionally. The rings are locked unconditionally from socket stacktraces. 

Trylocks are mostly used for non-critical periodic operations that can be lost and repeated later, or for operations that can be postponed (buffers return).

Exception is ring polling methods and I would prefer to change it. In general, trylocks is antipattern.

src/core/config/descriptor_providers/xlio_config_schema.json
pasis: Then need to improve the script. Multiple config parameter names were not converted to links by the script and I added them manually to the md file. Maybe we can make implicit format for tagging the parameters? Let's take it separately, sounds like an FR.

src/core/config/descriptor_providers/xlio_config_schema.json
pasis: probably some automated sync issue, will add back

pasis: actually the schema has a single \n across the file, so keeping single \n here makes it consistent

src/core/config/descriptor_providers/xlio_config_schema.json
pasis: fixed

src/core/config/descriptor_providers/xlio_config_schema.json
pasis: fixed

src/core/config/descriptor_providers/xlio_config_schema.json
pasis: fixed

docs/xlio_config_reference.md
```diff
@@ -2492,6 +2510,8 @@ delegate TCP timers mode (0).
 Consistently 0 = safe to disable. Consistently high = do not
 increase.
 
+**Pair with:** [[`periodic_drain_max_cqes`](#performancecompletion_queueperiodic_drain_max_cqes)](#performancecompletion_queueperiodic_drain_max_cqes) (either set to 0 disables).
```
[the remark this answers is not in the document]
pasis: fixed

src/core/config/descriptor_providers/xlio_config_schema.json
tomerdbz: sure, but at least for now let's align with the other usages. tick params with `  instead of putting markdown links

pasis: removed all the links from the schema and replaced them with full path parameters, so the script generates them propely.

docs/xlio_config_reference.md
```diff
@@ -3965,30 +4035,42 @@ is "delegate" (locks become no-ops).
...
   process networking inline during socket calls. Lowest
   latency, no extra CPU. Requires per-thread socket
   ownership and frequent socket calls.
-- *>0 (worker threads):* XLIO spawns N busy-polling
-  threads; operations dispatched via lock-free queues.
-  Sockets may be shared across application threads. One
-  listen socket suffices — XLIO creates one Receive
-  Side Scaling child per worker for parallel accept().
-  Outgoing connections distributed round-robin. Blocking
-  connect() is not supported.
```
tomerdbz, line 3982: blocking connect is not supported, don't we want to state that?

docs/xlio_config_reference.md
tomerdbz: wait, are you sure it only has effect in threads mode? doesn't seem so from sockinfo_tcp::rx_sleep_wait

docs/xlio_config_reference.md
```diff
@@ -2466,23 +2479,28 @@ consider increasing.
...
-updates, and keepalives. If the application has idle phases between
-socket calls, this timer prevents TCP stalls. If the application
-already drained the Completion Queue, the thread returns immediately
-(no redundant work).
+TCP requires execution context for ACKs and TCP progress. If the
+application has idle phases between socket calls, this timer prevents
+TCP stalls. If the application already drained the Completion Queue,
+the thread returns immediately (no redundant work).
+
+This feature guarantees TCP progress and avoids RQs overflow for
```
tomerdbz, line 2485: unless periodic_drain_max_cqes == 0, right?

docs/xlio_config_reference.md
```diff
@@ -3895,32 +3963,38 @@ lock overhead. Misuse symptoms: hung connections
...
-Uses epoll_wait with this value as minimum timeout.
+Base scheduling granularity (milliseconds) for XLIO's
+internal-thread event loop. Controls only the generic timer tick;
+protocol-specific periods are configured separately (TCP
+retransmission/keepalive/delayed-ACK cadence is set via
+[`network.protocols.tcp.timer_msec`](#networkprotocolstcptimer_msec)).
+
+Subsystems whose periodic/one-shot timers are paced by this resolution:
 
-**Constraint:** Lower bound for [`network.protocols.tcp.timer_msec`](#networkprotocolstcptimer_msec) — values
```
tomerdbz, line 3901: why remove this constraint?

docs/xlio_config_reference.md
```diff
@@ -3930,11 +4004,7 @@ and NGINX_DPU profiles use 32). Pair with a proportional
 >
 > **Maps to:** `XLIO_MULTILOCK`
 
-Selects spinlock or mutex for XLIO's ring, socket,
-TCP connection, and device locks. On the hot path
-XLIO uses non-blocking lock attempts (return immediately
-if contended), so the type matters mainly when contention
-forces a blocking wait.
+Selects spinlock or mutex for XLIO rings and sockets.
```
tomerdbz, line 4005: also TCP connection locks?

docs/xlio_config_reference.md
```diff
@@ -2540,7 +2565,7 @@ calls per second, even small values add measurable CPU overhead.
 Maximum TCP flows per ring using software Generic Receive Offload.
 Coalesces consecutive segments from the same flow into larger
 packets (up to 32 segments or ~64KB), reducing per-packet CPU
-overhead. Flows exceeding the limit bypass aggregation.
```
tomerdbz, line 2541: why remove "Flows exceeding the limit bypass aggregation."?

tomerdbz (changes requested):
more Qs and we're done

docs/xlio_config_reference.md
pasis: yes, it's used only by this flow:
```C
    // If the socket belongs to an entity_context read packets accordingly.
    if (m_entity_context) {
        return rx_read_ready_packets(p_iov, sz_iov, p_flags, __from, __fromlen, __msg);
    }
```

docs/xlio_config_reference.md
pasis: because it can be confusing for scenario of non-permanent connections. The config works similar to a pool. So we support this number of active GRO connections.

After second thought, the idea was good (to mention that extra connections continue working, just without GRO).

Will add it back.

docs/xlio_config_reference.md
pasis: It describes TCP timer behavior which is a separate subsystem and needs to be documented in the relevant section. There is a link below to the tcp timer config where it should be decribed.

docs/xlio_config_reference.md
pasis: socket locks include the TCP connection lock which locks most of the tcp socket operations. Besides this tcp specific lock, sockinfo object has a receive multilock. The rest socket locks are mutexes or spinlocks without multilock

docs/xlio_config_reference.md
pasis: Added non-blocking socket only to the limitations below.

docs/xlio_config_reference.md
pasis: there "Pair with" description below. It should imply that periodic_drain_max_cqes disables the feature (where feature means progress engine, not only the config).

Do you think we need to mention periodic_drain_max_cqes here as well?

tomerdbz: Not sure. I'm with you on this

tomerdbz (approved):

PR #627, opened 2026-05-28
issue: 4930789 Improve TCP RTT/RTO precision with microsecond timing

tomerdbz:
tcp_ticks advances only on the TCP slow-timer
(default ~200 ms), so RTT samples are frequently zero and, for any
sub-tick round trip the estimator cannot resolve, RTO sits at its
three-tick (600 ms) floor.

This commit implements the agreed design: keep the slow/fast
timer scheduling on ticks, but track RTT samples, the estimator state
(sa_us, sv_us), and the RTO deadline in monotonic microseconds.

  - New tcp_rto.{c,h} holds the pure VJ math (clamp, deadline_elapsed,
    backoff via the lwIP tcp_backoff[] shift table, and the Karn
    failed-send clear primitive) so the estimator is unit-testable
    without TCP. RTO backoff preserves the backoff shape (modulo the
    inherent +-1-tick quantization of the old `rtime >= rto` check,
    eliminated now that expiration is a true monotonic deadline): per
    retransmission rto_us = ((sa_us>>3)+sv_us) << tcp_backoff[nrtx],
    re-derived from the current estimator state on each pass and
    clamped to TCP_RTO_MAX_US (120 s).
  - TX records the sample start with clock_gettime_monotonic_us(); RX
    poll/drain sites refresh a per-thread g_xlio_tls_now_us once per
    non-empty batch (and once when GRO is holding aggregated packets
    about to flush).
  - tcp_output_segment() no longer bumps pcb->rtime /
    pcb->ticks_since_data_sent away from -1 before ip_output(). On a
    failed send the marker fields stay at the "no flight" sentinel,
    so TCP_USER_TIMEOUT and fast-retransmit dup-ACK gating no longer
    see a fake "data was just sent" point. The success path arms the
    markers atomically with rto_deadline_us inside
    tcp_rto_timer_start_if_needed().
  - The ERR_WOULDBLOCK move-to-unacked-and-recover dance is removed:
    failed sends now leave the segment on unsent for the next
    tcp_output() retry, so no fake unacked segment or armed RTO
    deadline is created.
  - getsockopt(TCP_INFO) returns rto_us directly for tcpi_rto, and
    populates tcpi_rtt/tcpi_rttvar from sa_us>>3 / sv_us>>2 when
    AC_CHECK_MEMBERS finds the fields.
  - The slow-timer bucket walk uses trylock; a new
    rto_check_consecutive_skips_max counter is exposed in socket_stats
    and aggregated into the tuning report so trylock starvation is
    observable in production.
  - OOSEQ retention preserves the legacy 3-tick wall-time floor
    (~3.6 s at default 200 ms cadence). Future work that wants the lower
    floor should re-propose it as a stand-alone change.

## Description

##### What

##### Why ?

##### How ?
architecture, approach, etc._

## Change type
- [ ] Bugfix
- [X] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list

## Summary by CodeRabbit

* **New Features**
  * Added configurable TCP retransmission-timeout floor from 1–120,000 ms, defaulting to 600 ms.
  * Improved RTT and retransmission timing precision using microsecond-based measurements.
  * Added blocking wait support for TCP send, receive, connect, and accept operations.
  * Added validation and support for 12-bit VLAN tags.
  * Added reporting for consecutive TCP timer-lock skips.

* **Documentation**
  * Clarified TCP timer, polling, worker-thread, and retransmission settings, including updated defaults and limits.


src/core/sock/sockinfo_tcp.cpp
```diff
@@ -4929,8 +4991,26 @@ void sockinfo_tcp::get_tcp_info(struct tcp_info *ti)
...
+     * via the standard >>3 / >>2 unpacking. Guarded by configure.ac so older
+     * toolchains without the fields still build. Each field is a single
+     * 4-byte aligned read; cross-field consistency is not guaranteed,
+     * matching the lock-free TCP_INFO semantics we already promise.
+     */
+#ifdef HAVE_STRUCT_TCP_INFO_TCPI_RTT
+    ti->tcpi_rtt = (uint32_t)(m_pcb.sa_us >> 3);
+#endif
+#ifdef HAVE_STRUCT_TCP_INFO_TCPI_RTTVAR
+    ti->tcpi_rttvar = (uint32_t)(m_pcb.sv_us >> 2);
```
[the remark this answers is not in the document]
tomerdbz: Verified empirically with kernel 5.15:

bare kernel BEFORE user data: tcpi_rttvar=192 us (NOT 0). The SYN→SYN-ACK RTT is a sample.
XLIO BEFORE user data: tcpi_rttvar=183 us, matches kernel within run-to-run noise.
The "pre-measurement returns 0" premise does not hold for a normally-handshaken socket on this kernel. XLIO already matches kernel for this case.

tomerdbz:
## PCB layout follow-up evidence (`3263e3ed`)

### Result

- Verdict: PASS
- Lab RED-GREEN: PR-head parent `1803eaed` versus exact layout candidate `3263e3ed`, ABBA repeated three times per workload.
- Performance: request-response throughput +0.0113% with cycles/op and client/server CPU neutral; streaming throughput neutral with instructions/transmitted-Gbit -5.9150%.
- Top risk: internal `tcp_pcb` offsets across compile-time CC and OOSEQ variants.

### What changed

This follow-up only reorders existing `tcp_pcb` fields and adds compile-time layout guards.
It adds no TCP behavior or runtime mechanism.
The default PCB falls from 424 to 416 bytes, and the four TX queue heads occupy one contiguous 64-byte intra-structure window.

### Local validation

- Five Jenkins build variants passed: debug, nginx-off, static-on, default, and envoy-on.
- Commit checks passed for `3263e3ed` and its parent.
- clang-format 16 style gate passed all 353 files.
- Unit tests passed 222/222.
- Hardware-free gtests passed 46/46, including `tcp_cc`.
- C syntax matrix passed for `TCP_CC_ALGO_MOD={0,1}` and `TCP_QUEUE_OOSEQ={0,1}`.
- Fresh maintainer review passed with no remaining findings.

### Lab E2E

- Suite: xlio_perf 1.0.0, source `af1e226f182358b5e71ca0ea3847d6864dedc2d5`
- Hosts: luna02 client and luna01 server
- Link: direct 200 Gb/s ConnectX-7
- Driver and OFED: mlx5_core `26.01-0.8.9`
- Firmware: client `28.42.1000`, server `32.43.2026`
- libxlio base: `1803eaed0ab4c86919e3ab62c331a1c1a9c236a8`
- libxlio candidate: `3263e3ed5f204e9f7e143d08d4341eb7009ddc24`
- Base static binary: `3d870322fcd344d9fee47de05bb1b52088e84362356f0bb3b624a6c6b68de579`
- Candidate static binary: `859556bf6d354c069d2c1b08a07ce539a909832a02ef584e22e7fd097ab53e58`
- Exact tested patch and committed diff SHA-256: `e1288fd156cdadeed995ca231fe860e553d1cbbfccf7f5dbe66b1b30f930acaf`
- XLIO config: lwIP CC, quick ACK, striding RQ 8192, LRO enabled, TSO enabled
- Request-response: 4 KiB, 32 connections, 16 client and 16 server workers, QD64, 50/50 read-write, per-connection, dedicated ports, 5-second warmup, 60-second measure
- Streaming: 32 KiB, 4,000 connections, 16 client and 16 server workers, QD16, client-send, dedicated ports, 5-second warmup, 60-second measure
- Sequence: 12 request-response runs and 12 streaming runs, six per arm for each workload
- Result: all 24 clients rc 0, zero failed/lost connections, zero client failures, zero streaming server failures, zero kernel error signals, and no NIC error/drop/discard/recovery growth
- Cleanup: no xlio_perf processes, no temporary aliases, and all hugepages free on both hosts

### Paired performance

Request-response:

- Throughput: +0.0113%, 95% CI [+0.0043%, +0.0193%]
- Client cores: +0.0268%, 95% CI [-0.0669%, +0.1000%]
- Cycles/op: +0.0218%, 95% CI [-0.0817%, +0.0813%]
- Instructions/op: +0.2720%, 95% CI [+0.0203%, +0.5823%]
- Server CPU ns/op: -0.0057%, 95% CI [-0.0335%, +0.0111%]

Streaming:

- Throughput: +0.0334%, 95% CI [-0.3045%, +0.4687%]
- Client cores: -0.0495%, 95% CI [-0.0873%, -0.0048%]
- Cycles/transmitted-Gbit: -0.0764%, 95% CI [-0.5300%, +0.2757%]
- Instructions/transmitted-Gbit: -5.9150%, 95% CI [-8.6350%, -5.6873%]
- Server CPU ns/byte: -0.0393%, 95% CI [-0.4618%, +0.3284%]

The request-response instruction increase is disclosed and did not produce a cycles, core-use, server-CPU, or throughput regression.
The request-response server JSON recorded non-fatal send/reset failures during disconnect teardown after successful client completion, 353 on base and 408 on candidate.

src/core/lwip/tcp.c
```diff
@@ -620,11 +622,15 @@ void tcp_slowtmr(struct tcp_pcb *pcb)
...
+        } else if (get_tcp_state(pcb) == SYN_SENT && pcb->nrtx >= TCP_SYNMAXRTX) {
             ++pcb_remove;
             err = ERR_TIMEOUT;
             LWIP_DEBUGF(TCP_DEBUG, ("tcp_slowtmr: max SYN retries reached\n"));
-        } else if (pcb->nrtx == TCP_MAXRTX) {
+            /* >= not ==: nrtx is incremented by several paths (fast retransmit,
+             * the gated probe, and RTO escalate), so it can step past the exact
+             * limit between two timer passes; an exact-equality abort could then
+             * be skipped until nrtx wrapped u8. */
+        } else if (pcb->nrtx >= TCP_MAXRTX) {
```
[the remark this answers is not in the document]
tomerdbz: Good one - will fix

src/stats/stats_printer.cpp
```diff
@@ -207,6 +207,11 @@ void print_full_stats(socket_stats_t *p_si_stats, mc_grp_info_t *p_mc_grp_info,
         fprintf(filename, "Retransmissions: %u\n", p_si_stats->counters.n_tx_retransmits);
     }
 
+    if (p_si_stats->n_tcp_timer_consecutive_skips_max) {
+        fprintf(filename, "TCP timer trylock skips (high-water): %u consecutive\n",
+                p_si_stats->n_tcp_timer_consecutive_skips_max);
+    }
```
pasis, line 213: I'm not sure we should expose this via stats and tuning report. I'd prefer removing trylocks (as separate feature) instead of putting effort into proper observability with documentation and then remove everything.

Now we know that the skips are not theoretical and you've collected numbers, so we can use this info to justify the trylock cleanup.

src/core/util/utils.h
pasis, line 38: the CACHELINE_SIZE is used only for the padding and the padding code can be C11. static_assert() is available either in assert.h or by C23 itself. The sizeof() can be changed to be C compatible.

[suggestion] Maybe move entire padding and asserts to some padding related header instead of moving only the cache line?

src/core/proto/xlio_time.h
```diff
@@ -0,0 +1,222 @@
+/*
+ * SPDX-FileCopyrightText: NVIDIA CORPORATION & AFFILIATES
+ * Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: GPL-2.0-only or BSD-2-Clause
+ */
+
+#ifndef XLIO_TIME_H_
+#define XLIO_TIME_H_
```
pasis, line 8: Same as xlio_cache.h: maybe move it to util/?

pasis (commented):
review is still in progress

src/core/proto/xlio_time.h
```diff
@@ -0,0 +1,222 @@
...
+ * See the RM#4930789 microsecond RTT/RTO design.
+ */
+extern __thread int64_t g_xlio_tls_now_us;
+
+/* --- Debug/test-build clock-read counters (design invariant I2) ---
+ *
+ * I2: at most one monotonic-clock read per non-empty RX batch, zero on empty
+ * polls, one per non-empty timer bucket pass; the defensive direct-clock
+ * fallbacks stay at zero. These per-thread counters make that invariant
+ * observable so unit tests, gtest, and lab runs can assert it directly.
```
pasis, line 43: [suggestion] Remove tags like I2. In the current procedures, design documents don't evolve when we change the code and there is no pointer to the design document.

src/core/proto/xlio_time.h
```diff
@@ -0,0 +1,222 @@
...
+    struct timespec ts;
+    int rc = clock_gettime(CLOCK_MONOTONIC, &ts);
+    assert(rc == 0);
+    (void)rc;
+    /* Suffix the literals to make the int64_t-wide multiply explicit
+     * regardless of where the cast binds. tv_sec is time_t (typically
+     * int64_t on 64-bit ABIs but only int32_t on Y2038-vulnerable 32-bit
+     * ABIs); the LL suffix forces 64-bit math even if a future refactor
+     * relocates the cast.
+     */
```
pasis, line 150: remove the comment. It's obvious from the code.

src/core/proto/xlio_time.h
```diff
@@ -0,0 +1,222 @@
...
+
+/* vDSO-inlined monotonic microsecond clock. Every TU that includes this
+ * header inlines the call site. The clock_gettime rc check is a standard
+ * assert(): checked when NDEBUG is undefined, DCE'd otherwise. No bespoke
+ * XLIO_TIME_* build flag.
+ */
+static inline int64_t clock_gettime_monotonic_us(void)
+{
+    struct timespec ts;
+    int rc = clock_gettime(CLOCK_MONOTONIC, &ts);
```
pasis, line 142: [question] Any specific reason to use clock_gettime() instead of c++ steady_clock? Is it required in C code?

src/core/lwip/tcp_rto.h
```diff
@@ -0,0 +1,233 @@
...
+_Static_assert(TCP_RTO_FLOOR_DEFAULT_US <= TCP_RTO_MAX_US,
+               "TCP_RTO_FLOOR_DEFAULT_US must not exceed TCP_RTO_MAX_US");
+_Static_assert(TCP_RTO_INITIAL_US >= TCP_RTO_FLOOR_DEFAULT_US,
+               "TCP_RTO_INITIAL_US must not undershoot TCP_RTO_FLOOR_DEFAULT_US");
+_Static_assert(TCP_RTO_INITIAL_US <= TCP_RTO_MAX_US,
+               "TCP_RTO_INITIAL_US must not exceed TCP_RTO_MAX_US");
+_Static_assert(TCP_RTO_FALLBACK_US >= TCP_RTO_FLOOR_DEFAULT_US,
+               "TCP_RTO_FALLBACK_US must not undershoot TCP_RTO_FLOOR_DEFAULT_US");
+_Static_assert(TCP_RTO_FALLBACK_US <= TCP_RTO_MAX_US,
+               "TCP_RTO_FALLBACK_US must not exceed TCP_RTO_MAX_US");
```
pasis, line 91: assert.h provides `static_assert()` for C and should handle it via wrapper or implementation based on the C standard.

src/core/lwip/tcp_rto.h
```diff
@@ -0,0 +1,233 @@
...
+
+/* Seed/reset the RTT/RTO estimator and deadline state on a PCB. Initial RTO
+ * comes from RFC 6298 (1 s), clamped by the configured floor. sa_us == 0 is
+ * the "no prior sample" sentinel,
+ * so sv_us is seeded to the initial RTO (matches the legacy lwIP seed
+ * shape) and rto_us is set explicitly to skip the estimator until the
+ * first valid sample arrives. Single source of truth used by tcp_pcb_init,
+ * tcp_pcb_recycle, and the unit-test PCB reset helper.
+ */
+void tcp_rto_pcb_seed(struct tcp_pcb *pcb);
```
pasis, line 227: [suggestion] Rename to _init() or _reset()

src/core/lwip/tcp_out.c
```diff
@@ -1319,17 +1307,13 @@ err_t tcp_output(struct tcp_pcb *pcb)
...
             rc = tcp_output_segment(seg, pcb);
-            if (rc != ERR_OK && pcb->unacked) {
-                /* Transmission failed, skip moving the segment to unacked, so we
-                 * retry with the next tcp_output(). We must have at least one unacked
-                 * segment in this case or RTO would be broken otherwise. */
+            if (rc != ERR_OK) {
+                /* Transmission failed before the segment reached the wire. Keep it
+                 * on unsent so the next tcp_output() retry does not depend on a
+                 * fake unacked segment or an armed RTO deadline.
+                 */
```
pasis, line 1314: the "fake unacked" segment refers to the previous implementation which is not visible anymore. Instead, describe current new behavior: keep the segment in the unsent and retry on the next tcp_output(), incoming ACK or <the new timed retry mechanism>.

src/core/lwip/tcp_out.c
```diff
@@ -1507,7 +1494,32 @@ static err_t tcp_output_segment(struct tcp_seg *seg, struct tcp_pcb *pcb)
...
+
+    if (rc == ERR_OK) {
+        /* Arm rto_deadline_us only on success and only if currently 0.
+         * Reuse tx_now_us from the sample start; otherwise read it now (cold path - first segment
+         * after idle).
+         */
+        if (pcb->rto_deadline_us == 0) {
+            if (!sample_started_here) {
+                tx_now_us = clock_gettime_monotonic_us();
+                xlio_time_dbg_inc_tx_arm_idle();
```
pasis, line 1516: [question] is it expected to debug "arm_idle" only if rtt sample wasn't started during this call?

src/core/lwip/tcp.c
```diff
@@ -109,21 +110,21 @@ const u8_t tcp_persist_backoff[7] = {3, 6, 12, 24, 48, 96, 120};
...
  * Called periodically to dispatch TCP timers.
  *
  */
-void tcp_tmr(struct tcp_pcb *pcb)
+void tcp_tmr(struct tcp_pcb *pcb, int64_t timer_now_us)
 {
     /* Call tcp_fasttmr() every (slow_tmr_interval / 2) ms */
     tcp_fasttmr(pcb);
 
     if (++(pcb->tcp_timer) & 1) {
```
pasis, line 124: here the slow timer is hardcoded as every other run, however, in set_tmr_resolution() you use TCP_SLOW_INTERVAL_FACTOR which is a (compile-time) parameter. If someone changes the parameter, this will lead to inconsistency.

src/core/lwip/tcp.c
```diff
@@ -644,27 +647,30 @@ void tcp_slowtmr(struct tcp_pcb *pcb)
...
                     }
 
-                    /* RFC 6298 5.5: Exponential backoff of the retransmission timer. */
-                    pcb->rto =
-                        tcp_clamp_rto_ticks(tcp_clamp_rto_signed_ticks((pcb->sa >> 3) + pcb->sv)
-                                            << tcp_backoff[pcb->nrtx]);
+                    /* RFC 6298 5.5: double the operational RTO. Keep this
+                     * independent of nrtx: an advancing ACK resets the retry
+                     * episode but a sample-less ACK retains this backed-off
+                     * value until a valid RTT sample recomputes it. */
```
pasis, line 669: Comment current new behavior or just refer to backoff algorithm. Don't mention old removed behavior (remove nrtx mention).

src/core/lwip/tcp.c
```diff
@@ -686,11 +692,18 @@ void tcp_slowtmr(struct tcp_pcb *pcb)
 
                     /* The following needs to be called AFTER cwnd is set to one mss */
                     tcp_rexmit_rto(pcb);
-                } else if (pcb->unacked == NULL && pcb->rtime >= 0 && pcb->unsent != NULL) {
-                    /* Retry tcp_output() when unsent data is stalled with an empty unacked queue.
-                     * This happens after fast retransmit empties unacked and tcp_output() fails
-                     * to send (e.g. segment split blocked by pbuf ref > 1). Without unacked
-                     * segments, the RTO above cannot fire and TCP progress can stop.
+                } else if (pcb->unacked == NULL && pcb->unsent != NULL) {
```
pasis, line 695: This condition can also mean zero receive window announced by the peer. Did you check that this doesn't interfere with the zero window probes?

2nd comment: [nit] the previous branch check for rto dedline elapsed and `unacked != NULL` is not obvious from the function name, so reader can assume that RTO is fired regardless of unacked state and we miss tcp_output() retry is `unacked == NULL`. Maybe move `unacked != NULL` check outside the rto deadline function?

src/core/lwip/tcp_rto.c
```diff
@@ -0,0 +1,230 @@
...
+     * SYN (zero length), whose SYN->SYN-ACK RTT seeds only the initial RTO and
+     * is closed at establishment. This excludes FIN-only / control-only
+     * segments from starting a sample. */
+    return pcb->rttest_us == 0 && !TCP_SEQ_LT(seg_seqno, pcb->snd_nxt) &&
+        (seg_len > 0 || seg_is_syn);
+}
+
+bool tcp_rto_deadline_elapsed(const struct tcp_pcb *pcb, int64_t timer_now_us)
+{
+    return pcb->unacked != NULL && pcb->rto_deadline_us != 0 &&
```
pasis, line 151: [nit] pcb->unacked check is not expected from the function name. Should it be moved outside?

src/core/lwip/tcp.h
```diff
@@ -306,24 +309,30 @@ struct tcp_pcb {
...
 #if TCP_CC_ALGO_MOD
     u32_t t_rttupdated; /* number of RTT estimations taken so far */
 #endif
-    s16_t sa, sv; /* @todo document this */
+    s32_t sa_us; /* 8 * SRTT in microseconds; 0 = "no prior sample" sentinel */
+    s32_t sv_us; /* 4 * RTTVAR in microseconds */
 
-    s16_t rto; /* retransmission time-out */
-    u8_t nrtx; /* number of retransmissions */
+    s32_t rto_us; /* retransmission timeout in microseconds */
```
pasis, line 331: why is it signed type? can it be negative?

src/core/lwip/tcp.c
```diff
@@ -744,12 +757,18 @@ void tcp_slowtmr(struct tcp_pcb *pcb)
...
+           microseconds and preserves the legacy 3-tick floor via
+           tcp_ooseq_timeout_us(). */
 #if TCP_QUEUE_OOSEQ
-        if (pcb->ooseq != NULL && (u32_t)tcp_ticks - pcb->tmr >= pcb->rto * TCP_OOSEQ_TIMEOUT) {
-            tcp_segs_free(pcb, pcb->ooseq);
-            pcb->ooseq = NULL;
-            LWIP_DEBUGF(TCP_CWND_DEBUG, ("tcp_slowtmr: dropping OOSEQ queued data\n"));
+        if (pcb->ooseq != NULL) {
+            uint64_t slow_interval_us = (uint64_t)slow_tmr_interval * 1000U;
+            uint64_t ooseq_idle_us = ((uint64_t)(u32_t)(tcp_ticks - pcb->tmr)) * slow_interval_us;
```
pasis, line 766: just cast to (uint64_t) without u32_t

src/core/lwip/tcp.c
```diff
@@ -744,12 +757,18 @@ void tcp_slowtmr(struct tcp_pcb *pcb)
 
         /* If this PCB has queued out of sequence data, but has been
            inactive for too long, will drop the data (it will eventually
-           be retransmitted). */
+           be retransmitted). The OOSEQ retention budget is now in
+           microseconds and preserves the legacy 3-tick floor via
+           tcp_ooseq_timeout_us(). */
```
pasis, line 762: avoid comparison to old implementation. Keep only actual description. The suffix _us already implies that the timeout is in us.

src/core/lwip/tcp_in.c
```diff
@@ -95,22 +100,71 @@ s8_t tcp_quickack(struct tcp_pcb *pcb, tcp_in_data *in_data)
 #endif
 }
 
-static bool tcp_handle_syn_established(struct tcp_pcb *pcb)
+/* Runs once per connection (closes the SYN->SYN-ACK RTT sample and seeds the
+ * initial RTO), but is called from the per-packet tcp_process(). Inlining it
+ * here bloats tcp_process()'s steady-state instruction footprint, which lowers
+ * IPC and connection-establishment rate on small-I-cache cores (measured ~7%
+ * CPS regression at 0KB Connection: close on BlueField-3). noinline keeps the
+ * seed code out-of-line and restores CPS to baseline; behavior is unchanged. */
```
pasis, line 108: Remove mention about baseline and that behavior is untouched

src/core/lwip/tcp_in.c
```diff
@@ -95,22 +100,71 @@ s8_t tcp_quickack(struct tcp_pcb *pcb, tcp_in_data *in_data)
...
+        } else if (now_us == 0) {
+            /* Cold TLS: the handshake seed reached tcp_process() without a
+             * refreshed RX timestamp. Unreachable today - every path into
+             * tcp_process() routes through the guarded CQ refresh hooks - and
+             * kept as defense in depth: skipping the seed leaves the safe
+             * RFC-compliant 1 s initial RTO, whereas seeding unconditionally
+             * would feed a negative handshake delta and clamp garbage into
+             * rto_us, hiding a future RX-refresh coverage regression.
+             * Contract I3 requires a missed refresh to be loud in dev builds,
+             * matching the SYN-SENT rearm fallback in tcp_process() below. */
```
pasis, line 155: This is simply mustn't be reachable, or unacceptable as the argument value. Reduce the comment. now_us is an argument and such function shouldn't rely on knowledge how it's obtained and maintained. This can introduce risks to leave such code stale when the design of the timestamps is changed.

src/core/lwip/tcp_in.c
```diff
@@ -95,22 +100,71 @@ s8_t tcp_quickack(struct tcp_pcb *pcb, tcp_in_data *in_data)
...
+         * a clean first sample. In particular, do not copy a configured floor
+         * above 1 s into sv_us: that would masquerade as the conservative
+         * SYN-loss fallback and transiently inflate TCP_INFO.tcpi_rttvar.
+         * now_us == 0 means cold TLS. now_us <= rttest_us means that this
+         * packet reused a batch timestamp captured before an in-batch send.
+         * Neither timestamp can yield a causal RTT, so skip the seed but still
+         * close the sample. */
+        if (now_us > pcb->rttest_us) {
+            const s32_t seed_us = tcp_rto_seed_from_handshake_us(now_us - pcb->rttest_us);
+            pcb->rto_us = seed_us;
```
pasis, line 145: merge the 2 lines

src/core/lwip/tcp_in.c
```diff
@@ -1144,11 +1224,44 @@ static void tcp_receive(struct tcp_pcb *pcb, tcp_in_data *in_data)
                 pcb->flags &= ~TF_INFR;
             }
 
-            /* Reset the number of retransmissions. */
+            /* Forward progress starts a new retry episode. This does not
+             * collapse RTO backoff: rto_us holds the operational value and is
+             * recomputed only when a valid RTT sample is available below. */
```
pasis, line 1229: remove mention of the old implementation. nrtx is not related to backoff anymore.

src/core/lwip/tcp_in.c
```diff
@@ -1189,20 +1302,69 @@ static void tcp_receive(struct tcp_pcb *pcb, tcp_in_data *in_data)
...
+                        LWIP_DEBUGF(TCP_RTO_DEBUG,
+                                    ("tcp_receive: rearming RTO with direct clock, "
+                                     "TLS now_us not refreshed\n"));
+                        assert(now_us_valid &&
+                               "partial ACK rearm reached tcp_receive without refreshed RX "
+                               "timestamp");
+                        rearm_now_us = clock_gettime_monotonic_us();
+                        xlio_time_dbg_inc_fallback();
+                    }
+                    tcp_rto_timer_rearm(pcb, rearm_now_us);
```
pasis, line 1366: Do unconditional RTO re-arm and remove it from tcp_rtt_estimator_update_and_rearm_us() - this will improve readability.

src/core/lwip/tcp_in.c
```diff
@@ -1189,20 +1302,69 @@ static void tcp_receive(struct tcp_pcb *pcb, tcp_in_data *in_data)
...
-                pcb->ticks_since_data_sent = 0;
+                if (rtt_sample_eligible) {
+#if TCP_CC_ALGO_MOD
+                    pcb->t_rttupdated++;
+#endif
+                    LWIP_DEBUGF(TCP_RTO_DEBUG,
+                                ("tcp_receive: experienced rtt %lld us\n",
+                                 (long long)(ack_now_us - pcb->rttest_us)));
+
+                    tcp_rtt_estimator_update_and_rearm_us(pcb, ack_now_us, ack_now_us);
```
pasis, line 1349: [nit] tcp_rtt_estimator_update_and_rearm_us() accepts a timestamp while tcp_rtt_estimator_update_us() accepts some processed raw_sample_us - this is inconsistency in the interfaces.

src/core/lwip/tcp_in.c
```diff
@@ -1189,20 +1302,69 @@ static void tcp_receive(struct tcp_pcb *pcb, tcp_in_data *in_data)
...
                     pcb->persist_cnt = 0;
                     pcb->persist_backoff = 1;
                 }
-                pcb->rtime = -1;
-                pcb->ticks_since_data_sent = -1;
+                tcp_rto_timer_stop(pcb);
             } else {
-                pcb->rtime = 0;
-                pcb->ticks_since_data_sent = 0;
+                if (rtt_sample_eligible) {
```
pasis, line 1341: Can this copy-paste block be unconditional above? update RTT regardless of unacked state. Re-arm RTT sample in the partial ACK branch.

src/stats/stats_printer.cpp
tomerdbz: Not sure, the cost is not big - and I haven't run on a full matrix. I'd really like to know how/if it manifests in ver. I can remove it - just want you to give an extra thought before we do as it has dramatic effects on RTO. (one more timer tick, no accounting)

tomerdbz: maybe just keep the log warn? so at least we have basic visibility?

src/core/util/xlio_time.h
```diff
@@ -0,0 +1,222 @@
+/*
+ * SPDX-FileCopyrightText: NVIDIA CORPORATION & AFFILIATES
+ * Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: GPL-2.0-only or BSD-2-Clause
+ */
+
+#ifndef XLIO_TIME_H_
+#define XLIO_TIME_H_
```
tomerdbz: done

src/core/proto/xlio_time.h
tomerdbz: done

src/core/proto/xlio_time.h
tomerdbz: done

src/core/util/xlio_time.h
```diff
@@ -0,0 +1,222 @@
...
+
+/* vDSO-inlined monotonic microsecond clock. Every TU that includes this
+ * header inlines the call site. The clock_gettime rc check is a standard
+ * assert(): checked when NDEBUG is undefined, DCE'd otherwise. No bespoke
+ * XLIO_TIME_* build flag.
+ */
+static inline int64_t clock_gettime_monotonic_us(void)
+{
+    struct timespec ts;
+    int rc = clock_gettime(CLOCK_MONOTONIC, &ts);
```
tomerdbz: it is consumed by the lwIP C translation units (tcp.c, tcp_in.c, tcp_out.c, tcp_rto.c), where std::chrono is not available.

src/core/lwip/tcp_rto.h
tomerdbz: fixed

src/core/lwip/tcp_out.c
tomerdbz: probably will revert this anyway - H1 shows this was a more delicate change than it seemed. out-of-scope/time to fix that

tomerdbz: will make sure comments either wayt are not hisotry lessooons

src/core/util/utils.h
tomerdbz: applied in the cleanup commit: the C/C++ static_assert dual blocks are gone and xlio_cache.h moved to util/ next to xlio_time.h.
I did not move PADDING()/BOUNDARY_SIZE_ASSERT in this PR: the tcp_pcb layout asserts are offset-range asserts, not the `__explicit_padding` boundary pattern, so the lwIP code would not use those macros even if they were C11-compatible.
Let's do the padding-machinery consolidation (including the C-compatible `sizeof` form) as a separate small cleanup.

src/core/lwip/tcp_rto.h
tomerdbz: done

src/core/lwip/tcp.c
tomerdbz: Fixed via a compile-time pin: `static_assert(TCP_SLOW_INTERVAL_FACTOR == 2, ...)` now sits directly at the alternation in `tcp_tmr()`, so changing the factor becomes a build error at the exact spot instead of a silent inconsistency.

src/core/lwip/tcp_out.c
tomerdbz: Yes, exactly. note it's optinal observability (and you can see how it's used for unit tests) to ensure correctness in debug builds.

src/core/lwip/tcp.c
tomerdbz: Done

src/core/lwip/tcp_rto.c
tomerdbz: Done

src/core/lwip/tcp.h
tomerdbz: It is `s32_t` to stay in the signed domain of the surrounding arithmetic (`sa_us`/`sv_us` are `s32_t`, deadlines are `int64_t`, e.g. `(int64_t)pcb->rto_us << 1`); unsigned would force sign-conversion casts at every int64 mixing site for no gain.
Can switch to `u32_t` if you feel strongly, but I think signed is the lower-friction choice here.

src/core/lwip/tcp.c
tomerdbz: Done

src/core/lwip/tcp.c
tomerdbz: Done

src/core/lwip/tcp_in.c
tomerdbz: Done

src/core/lwip/tcp_in.c
tomerdbz: Done - the block now just states that `now_us == 0` is not an acceptable value here (assert + defensive skip of the seed), without any narrative about how the caller obtains or maintains the timestamp.

src/core/lwip/tcp_in.c
tomerdbz: Done

src/core/lwip/tcp_in.c
tomerdbz: Done

src/core/lwip/tcp_in.c
tomerdbz: Done. the composite helper is gone, so only `tcp_rtt_estimator_update_us(pcb, raw_sample_us)` remains.

src/core/lwip/tcp_in.c
tomerdbz: Done - the estimator update now runs once, above the unacked split, for both full and partial ACKs.

src/core/lwip/tcp.c
tomerdbz: reverting this change and this risk will be reverted - not because of the zero window probes (I can test it and believe it's fine) but because H1 shows this was a more delicate change than it seemed. out-of-scope/time to fix that (really unrelated to the PR. though I strongly feel the legacy fake move to unacked is bad)

src/stats/stats_printer.cpp
pasis: Let's keep trylocks removal for the next step.

src/core/lwip/tcp_out.c
```diff
@@ -1319,15 +1329,20 @@ err_t tcp_output(struct tcp_pcb *pcb)
             }
 
             rc = tcp_output_segment(seg, pcb);
-            if (rc != ERR_OK && pcb->unacked) {
-                /* Transmission failed, skip moving the segment to unacked, so we
-                 * retry with the next tcp_output(). We must have at least one unacked
-                 * segment in this case or RTO would be broken otherwise. */
+            if (rc != ERR_OK && (pcb->unacked != NULL || TCP_SEGLEN(seg) == 0)) {
```
pasis, line 1332: TCP_SEGLEN(seg) must be always > 0 for the queued segment objects. And we don't have any special features that would keep empty object in the queue,

tomerdbz:
bot:retest

pasis (approved):
There are followup items that need to be addressed in a separate PR

tomerdbz:
followups:
Remove C-style _Static_assert() copypaste and use assert.h static_assert() for both c/c++
Remove tcp trylocks and their accounting/stats/documentation
PADDING()/BOUNDARY_SIZE_ASSERT cleanup if needed (make it C11-compatible and maybe move to some header like the cache line size)
TCP_SLOW_INTERVAL_FACTOR  assert to be 2 is not there as claimed. This leads to potential inconsistency between hardcoded 2 in tcp_tmr() and parametrized value in other places
if (rc != ERR_OK && (pcb->unacked != NULL || TCP_SEGLEN(seg) == 0)) { --- segment object length cannot be 0 in tcp_output()

tomerdbz:
bot:retest

PR #649, opened 2026-08-18
issue: 5226259 Add openssl development files to build dependencies

pasis:
## Description

TLS offload support is gated on openssl/evp.h in config/m4/utls.m4. With the default --enable-utls=auto, the build system silently disables UTLS support if the openssl header is missing.

Neither the rpm spec nor debian/control required the openssl development files, so the default binary packages were built without TLS offload whenever the building setup missed openssl headers.

Add openssl-devel to BuildRequires in libxlio.spec.in and libssl-dev to Build-Depends in debian/control. This is a build-time dependency only.

Also install the package in the CI images that compile XLIO, so the new build dependency resolves and the TLS code is actually covered by the build, static analysis and packaging jobs. Dockerfile.ol9.4 already had it.

##### What
Add openssl-devel to BuildRequires in libxlio.spec.in and libssl-dev to Build-Depends in debian/control.

##### Why ?
Build libxlio binary packages with TLS support explicitly.

## Change type
What kind of change does this PR introduce?
- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactoring (no functional changes, no api changes)
- [x] Build related changes
- [x] CI related changes
- [ ] Documentation content changes
- [ ] Tests
- [ ] Other

## Check list
- [ ] Code follows the style de facto guidelines of this project
- [ ] Comments have been inserted in hard to understand places
- [ ] Documentation has been updated (if necessary)
- [ ] Test has been added (if possible)

## Summary by CodeRabbit

* **Chores**
  * Updated supported build and packaging environments with required OpenSSL development dependencies.
  * Improved consistency of dependency configuration across RHEL, Ubuntu, Debian, and RPM-based builds.

PR #654, opened 2026-08-21
[CI] issue: INFINIOPS-922 Use internal ctyunOS repo mirror

pasis:
## Description

**What**: Rewrote the ctyunOS 23.01 repo setup to write
      /etc/yum.repos.d/ctyunos.repo whole, pointing everything at
      the internal webrepo.mtl.labs.mlnx mirror and keeping update
      and extras defined but disabled.
**How**: Replaced the three sed patches with a printf that emits the
     full repo file, plus a cat of the result so the effective repo
     config lands in the build log.
**Why**: The sed patches exit 0 when they do not match, so a base image
     layout change would silently leave unreachable repos enabled;
     the internal mirror serves the everything snapshot until the
     gtm mirror path is available again.

##### What
Update CtyunOS repo mirrors.

##### Why ?
Fix CI failures during image prepare.

## Change type
- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Build related changes
- [x] CI related changes
- [ ] Tests
- [ ] Other

## Check list

### DIFF
--- a/.ci/dockerfiles/Dockerfile.ctyunos23.01
+++ b/.ci/dockerfiles/Dockerfile.ctyunos23.01
@@ -7,9 +7,36 @@ ARG _GID=101
 ARG _LOGIN=swx-jenkins
 ARG _HOME=/var/home/$_LOGIN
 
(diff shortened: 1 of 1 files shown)


pasis:
Merged as part of #649.

PR #656, opened 2026-08-31
issue: 5094361 Enable relaxed ordering for XLIO memory registrations

BasharRadya:
Request IBV_ACCESS_RELAXED_ORDERING for XLIO-owned host memory registrations so per_mkey mode can improve performance without a device-wide mlxconfig change.

Keep the runtime switch disabled by default to preserve existing ordering behavior. Older verbs builds continue with regular ordering, while modern libibverbs treats the flag as optional when kernel support is unavailable.

Document both configuration paths and their user-visible behavior.

## Description

##### What

##### Why ?

##### How ?
architecture, approach, etc._

## Change type
- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list

## Summary by CodeRabbit

* **New Features**
  * Added an optional relaxed-ordering setting for XLIO-managed memory registration.
  * Configure it through `hardware_features.memory_registration.relaxed_ordering` or `XLIO_ENABLE_RELAXED_ORDERING`.
  * The setting is disabled by default and applies only when supported by the system.

* **Documentation**
  * Updated the configuration reference with the new setting, default behavior, scope, and support details.

* **Diagnostics**
  * Startup output now reports the relaxed-ordering configuration and memory-registration logs include access flags.


src/core/config/mappings.py
```diff
@@ -44,6 +44,7 @@
     "network.timing.hw_ts_conversion": "XLIO_HW_TS_CONVERSION",
     
     # hardware_features section
+    "hardware_features.memory_registration.relaxed_ordering": "XLIO_ENABLE_RELAXED_ORDERING",
```
pasis: [comment] please check the convention in other names to follow it. This can be ENABLE_RELAXED_ORDERING or RELAXED_ORDERING_ENABLE, or even just RELAXED_ORDERING. Just to keep it consistent. If the convention is different, adjust other names too, such as CONFIG_VAR_...

docs/xlio_config_reference.md
```diff
@@ -31,6 +31,7 @@ This file documents all 121 XLIO runtime configuration parameters with their typ
   - [`core.syscall.fork_support`](#coresyscallfork_support) — Enable fork support
   - [`core.syscall.sendfile_cache_limit`](#coresyscallsendfile_cache_limit) — Sendfile byte limit
 - **[HARDWARE_FEATURES](#hardware_features)**
+  - [`hardware_features.memory_registration.relaxed_ordering`](#hardware_featuresmemory_registrationrelaxed_ordering) — Enable relaxed ordering for memory registration
```
pasis: [comment] Please see whether we already have existing category related to memory registration, PCI or similar. Create new category only if the new config doesn't fit anything existing.

docs/xlio_config_reference.md
```diff
@@ -748,6 +749,33 @@ working set = 10GB → default 10GB is appropriate.
...
+does not support it, and builds against older verbs register memory without the flag.
+
+**Values:**
+
+- true: Request Relaxed Ordering for XLIO-owned host memory regions. This can improve
+  performance by allowing NIC DMA transfers to reach memory out of order. Completion semantics
+  and data visibility are unchanged, but intermediate write-after-write ordering is not guaranteed.
+- false (default): Register memory with the previous strict-ordering access flags. This preserves
+  historical XLIO behavior and can be used for compatibility, troubleshooting, or performance
+  comparison.
```
pasis: 'false' keeps the current HW configuration. If the HW is configured with force_relaxed, it doesn't force strict ordering AFAIU.

Also I would like to see an explicit note that this config parameter doesn't affect force_relaxed mode.

docs/xlio_config_reference.md
```diff
@@ -748,6 +749,33 @@ working set = 10GB → default 10GB is appropriate.
...
+> **Type:** boolean
+>
+> **Maps to:** `XLIO_ENABLE_RELAXED_ORDERING`
+
+Requests `IBV_ACCESS_RELAXED_ORDERING` for host memory regions registered by XLIO.
+
+When the NIC Relaxed Ordering mode is configured as `per_mkey`, this option lets XLIO request
+Relaxed Ordering for its own memory keys without changing the device-wide `mlxconfig` setting.
+The request is best-effort: libibverbs ignores the optional flag when the kernel or provider
+does not support it, and builds against older verbs register memory without the flag.
```
pasis: The param description contains low level details that can complicate reading and understanding by user. I'd suggest to write short high level description first and than extra notes about corner cases and support requirements.

Also the best effort part is concerning to me. I'd like to understand that the feature is not applied if I request it explicitly, at least with a log message, because this can affect performance numbers and decision making based on the numbers.

src/core/dev/ib_ctx_handler.cpp
```diff
@@ -400,6 +400,14 @@ uint32_t ib_ctx_handler::mem_reg(void *addr, size_t length, uint64_t access)
     struct ibv_mr *mr = nullptr;
     uint32_t lkey = LKEY_ERROR;
 
+    if (safe_mce_sys().enable_relaxed_ordering) {
+#ifndef DEFINED_IBV_ACCESS_RELAXED_ORDERING
+        VLOG_PRINTF_ONCE_THEN_DEBUG(
+            VLOG_WARNING,
+            "IBV_ACCESS_RELAXED_ORDERING is unavailable. Continuing without relaxed ordering\n");
+#endif
+        access |= XLIO_IBV_ACCESS_RELAXED_ORDERING;
```
pasis: should this line be under `#else` block?

src/core/ib/base/verbs_extra.h
```diff
@@ -170,6 +170,11 @@ typedef struct ibv_tso_caps xlio_ibv_tso_caps;
 typedef struct ibv_send_wr xlio_ibv_send_wr;
 // ibv_reg_mr
 #define XLIO_IBV_ACCESS_LOCAL_WRITE IBV_ACCESS_LOCAL_WRITE
+#ifdef DEFINED_IBV_ACCESS_RELAXED_ORDERING
+#define XLIO_IBV_ACCESS_RELAXED_ORDERING IBV_ACCESS_RELAXED_ORDERING
+#else
+#define XLIO_IBV_ACCESS_RELAXED_ORDERING 0
+#endif
```
pasis: I'd prefer to use IBV_ACCESS_RELAXED_ORDERING directly without hiding it behind a macro. We don't support experimental verbs anymore and can do cleanup of some xlio_ prefixed definitions.

src/core/util/sys_vars.cpp
```diff
@@ -815,6 +815,7 @@ void mce_sys_var::get_env_params()
     strq_stride_num_per_rwqe = MCE_DEFAULT_STRQ_NUM_STRIDES;
     strq_stride_size_bytes = MCE_DEFAULT_STRQ_STRIDE_SIZE_BYTES;
     strq_strides_compensation_level = MCE_DEFAULT_STRQ_STRIDES_COMPENSATION_LEVEL;
+    enable_relaxed_ordering = MCE_DEFAULT_ENABLE_RELAXED_ORDERING;
```
pasis: [no action needed] I'd prefer to define a convention in the future to have prefix in the name first and action part after it. For example: `strq_enable`, `relaxed_ordering_enable` (or `relaxed_ordering_force`). In this way all the strq related parameters have the same prefix and grouped visually.

src/core/main.cpp
```diff
@@ -616,6 +616,9 @@ void print_env_vars_xlio_global_settings()
     VLOG_STR_PARAM_STRING("Striding RQ", option_3::to_str(safe_mce_sys().enable_strq_env),
                           option_3::to_str(MCE_DEFAULT_STRQ), SYS_VAR_STRQ,
                           option_3::to_str(safe_mce_sys().enable_strq_env));
+    VLOG_PARAM_STRING("Relaxed Ordering", safe_mce_sys().enable_relaxed_ordering,
+                      MCE_DEFAULT_ENABLE_RELAXED_ORDERING, SYS_VAR_ENABLE_RELAXED_ORDERING,
+                      safe_mce_sys().enable_relaxed_ordering ? "Enabled" : "Disabled");
```
pasis: This text can be confusing. It's not necessary disabled with 'false', because force_relaxed mode will provide relaxed ordering while XLIO header says it's "disabled".

pasis (commented):
Looks good in general. Small changes needed.

src/core/config/mappings.py
BasharRadya: XLIO_ENABLE_RELAXED_ORDERING → XLIO_RELAXED_ORDERING
Renamed related SYS_VAR, CONFIG_VAR, and default constants accordingly.

docs/xlio_config_reference.md
BasharRadya: no related category, i can simply remove the memory_Registeration and put relaxed oredering underas hardware features
WDYT?

docs/xlio_config_reference.md
BasharRadya: Meaning of false

Docs now say XLIO does not request RO.
Explicitly clarify it does not force strict ordering.

BasharRadya: force_relaxed

Documented that hardware RO remains active regardless of the XLIO setting.

docs/xlio_config_reference.md
BasharRadya: Added a short user-level description first.
Moved technical details into hardware/support sections.
Added an explicit warning when the verbs flag is unavailable.
Kernel-side silent masking cannot be detected through the public verbs API, so docs clarify that logs report the request—not effective hardware state.

src/core/dev/ib_ctx_handler.cpp
BasharRadya: #else placement

Moved access |= IBV_ACCESS_RELAXED_ORDERING into the supported #else branch.

src/core/ib/base/verbs_extra.h
BasharRadya: Verbs wrapper macro

Removed XLIO_IBV_ACCESS_RELAXED_ORDERING.
Use IBV_ACCESS_RELAXED_ORDERING directly.

src/core/main.cpp
BasharRadya: Changed “Enabled/Disabled” to “Requested/Not requested”.

PR #657, opened 2026-09-01
Feature 4975332 worker mode tls hw support

BasharRadya:
worker mode tls hw support

## Change type
- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list

### DIFF
--- a/src/core/dev/hw_queue_tx.cpp
+++ b/src/core/dev/hw_queue_tx.cpp
@@ -877,15 +877,21 @@ std::unique_ptr<dpcp::tls_dek> hw_queue_tx::get_tls_dek(const void *key, uint32_
         return std::unique_ptr<dpcp::tls_dek>(nullptr);
     }
 
+    if (unlikely(!m_p_ring->tls_sync_dek_supported())) {
+        return get_new_tls_dek(key, key_size_bytes);
+    }
+
+    m_tls_cache_lock.lock();
+
     // If the amount of available DEKs in m_dek_put_cache is smaller than
     // low-watermark we continue to create new DEKs. This is to avoid situations
     // where one DEKs is returned and then fetched in a throttlling manner
     // causing too frequent crypto-sync.
     // It is also possible that crypto-sync may have higher impact with higher number
     // of active connections.
-    if (unlikely(!m_p_ring->tls_sync_dek_supported()) ||
-        (unlikely(m_tls_dek_get_cache.empty()) &&
-         (m_tls_dek_put_cache.size() <= safe_mce_sys().utls_low_wmark_dek_cache_size))) {
+    if (unlikely(m_tls_dek_get_cache.empty()) &&
+        (m_tls_dek_put_cache.size() <= safe_mce_sys().utls_low_wmark_dek_cache_size)) {
+        m_tls_cache_lock.unlock();
         return get_new_tls_dek(key, key_size_bytes);
     }
 
@@ -893,8 +899,10 @@ std::unique_ptr<dpcp::tls_dek> hw_queue_tx::get_tls_dek(const void *key, uint32_
         hwqtx_logdbg("Empty DEK get cache. Swapping caches and do Sync-Crypto. Put-Cache size: %zu",
                      m_tls_dek_put_cache.size());
 
+        /* Rare operation. Keep the cache lock so the check-sync-swap sequence stays atomic. */
         status = adapter->sync_crypto_tls();
         if (unlikely(status != dpcp::DPCP_OK)) {
+            m_tls_cache_lock.unlock();
             hwqtx_logwarn("Failed to flush DEK HW cache, status: %d", status);
             return get_new_tls_dek(key, key_size_bytes);
         }
@@ -904,7 +912,9 @@ std::unique_ptr<dpcp::tls_dek> hw_queue_tx::get_tls_dek(const void *key, uint32_
 
     std::unique_ptr<dpcp::tls_dek> out_dek(std::move(m_tls_dek_get_cache.front()));
     m_tls_dek_get_cache.pop_front();
+    m_tls_cache_lock.unlock();
 
+    /* Modify an exclusively owned object - outside the lock. */
     struct dpcp::dek_attr dek_attr;
     memset(&dek_attr, 0, sizeof(dek_attr));
     dek_attr.key_blob = const_cast<void *>(key);
@@ -926,42 +936,55 @@ void hw_queue_tx::put_tls_dek(std::unique_ptr<dpcp::tls_dek> &&tls_dek_obj)
         return;
     }
     // We don't allow unlimited DEK cache to avoid system DEK starvation.
-    if (likely(m_p_ring->tls_sync_dek_supported()) &&
-        m_tls_dek_put_cache.size() < safe_mce_sys().utls_high_wmark_dek_cache_size) {
-        m_tls_dek_put_cache.emplace_back(std::forward<std::unique_ptr<dpcp::tls_dek>>(tls_dek_obj));
+    if (likely(m_p_ring->tls_sync_dek_supported())) {
+        std::lock_guard<decltype(m_tls_cache_lock)> lock(m_tls_cache_lock);
+        if (m_tls_dek_put_cache.size() < safe_mce_sys().utls_high_wmark_dek_cache_size) {
+            m_tls_dek_put_cache.emplace_back(
+                std::forward<std::unique_ptr<dpcp::tls_dek>>(tls_dek_obj));
+        }
     }
 }
 
-xlio_tis *hw_queue_tx::tls_context_setup_tx(const xlio_tls_info *info)
+xlio_tis *hw_queue_tx::tls_reserve_tis(const xlio_tls_info *info)
 {
     std::unique_ptr<xlio_tis> tis;
-    if (m_tls_tis_cache.empty()) {
+
+    m_tls_cache_lock.lock();
+    if (!m_tls_tis_cache.empty()) {
+        tis.reset(m_tls_tis_cache.back());
+        m_tls_tis_cache.pop_back();
+    }
+    m_tls_cache_lock.unlock();
+
+    if (!tis) {
+        /* Firmware command on the adapter - no lock is required. */
         tis = create_tis(DPCP_TIS_FLAGS | dpcp::TIS_ATTR_TLS);
         if (unlikely(!tis)) {
             return nullptr;
         }
-    } else {
-        tis.reset(m_tls_tis_cache.back());
-        m_tls_tis_cache.pop_back();
     }
 
     auto dek_obj = get_tls_dek(info->key, info->key_len);
     if (unlikely(!dek_obj)) {
+        std::lock_guard<decltype(m_tls_cache_lock)> lock(m_tls_cache_lock);
         m_tls_tis_cache.push_back(tis.release());
         return nullptr;
     }
 
     tis->assign_dek(std::move(dek_obj));
+    return tis.release();
+}
+
+void hw_queue_tx::tls_context_setup_tx(const xlio_tls_info *info, xlio_tis *tis)
+{
     uint32_t tisn = tis->get_tisn();
 
-    tls_post_static_params_wqe(tis.get(), info, tisn, tis->get_dek_id(), 0, false, true);
-    tls_post_progress_params_wqe(tis.get(), tisn, 0, false, true);
+    tls_post_static_params_wqe(tis, info, tisn, tis->get_dek_id(), 0, false, true);
+    tls_post_progress_params_wqe(tis, tisn, 0, false, true);
     /* The 1st post after TLS configuration must be with fence. */
     m_b_fence_needed = true;
 
     assert(!tis->m_released);
-
-    return tis.release();
 }
 
 void hw_queue_tx::tls_context_resync_tx(const xlio_tls_info *info, xlio_tis *tis, bool skip_static)
@@ -975,36 +998,30 @@ void hw_queue_tx::tls_context_resync_tx(const xlio_tls_info *info, xlio_tis *tis
     m_b_fence_needed = true;
 }
 
-int hw_queue_tx::tls_context_setup_rx(xlio_tir *tir, const xlio_tls_info *info,
-                                      uint32_t next_record_tcp_sn, xlio_comp_cb_t callback,
-                                      void *callback_arg)
+int hw_queue_tx::tls_reserve_rx_dek(xlio_tir *tir, const xlio_tls_info *info)
 {
-    uint32_t tirn;
-    dpcp::tls_dek *_dek;
-    dpcp::status status;
-    dpcp::adapter *adapter = m_p_ib_ctx_handler->get_dpcp_adapter();
-    struct dpcp::dek_attr dek_attr;
-
-    memset(&dek_attr, 0, sizeof(dek_attr));
-    dek_attr.key_blob = (void *)info->key;
-    dek_attr.key_blob_size = info->key_len;
-    dek_attr.key_size = info->key_len;
-    dek_attr.pd_id = adapter->get_pd();
-    status = adapter->create_tls_dek(dek_attr, _dek);
-    if (unlikely(status != dpcp::DPCP_OK)) {
-        hwqtx_logerr("Failed to create DEK, status: %d", status);
+    /* RX DEKs are not recycled - create a fresh one. It is destroyed together with the TIR. */
+    auto dek_obj = get_new_tls_dek(info->key, info->key_len);
+    if (unlikely(!dek_obj)) {
+        hwqtx_logerr("Failed to create DEK");
         return -1;
     }
-    tir->assign_dek(_dek);
-    tir->assign_callback(callback, callback_arg);
-    tirn = tir->get_tirn();
+    tir->assign_dek(dek_obj.release());
+    return 0;
+}
 
-    tls_post_static_params_wqe(NULL, info, tirn, _dek->get_key_id(), 0, false, false);
+void hw_queue_tx::tls_context_setup_rx(xlio_tir *tir, const xlio_tls_info *info,
+                                       uint32_t next_record_tcp_sn, xlio_comp_cb_t callback,
+                                       void *callback_arg)
+{
+    uint32_t tirn = tir->get_tirn();
+
+    tir->assign_callback(callback, callback_arg);
+
+    tls_post_static_params_wqe(NULL, info, tirn, tir->get_dek_id(), 0, false, false);
     tls_post_progress_params_wqe(tir, tirn, next_record_tcp_sn, false, false);
 
     assert(!tir->m_released);
-
-    return 0;
 }
 
 void hw_queue_tx::tls_resync_rx(xlio_tir *tir, const xlio_tls_info *info, uint32_t hw_resync_tcp_sn)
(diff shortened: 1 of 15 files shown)


BasharRadya:
bot:retest

src/core/event/entity_context.h
```diff
@@ -50,20 +53,37 @@ class entity_context : public poll_group {
...
+        {
+        }
+
         job_type job_id;
         int flags;
         sockinfo *sock;
         mem_buf_desc_t *buf;
         uint32_t offset;
         uint32_t tot_size;
+        std::shared_ptr<const tx_call_ctx> tx_ctx;
```
AlexanderGrissik: This can be vary dangerous. if shared_ptr has atomic ops on initialization and move() even with nulled ptr.

src/core/sock/sockinfo.h
```diff
@@ -201,6 +203,35 @@ struct xlio_tx_call_attr_t {
...
+    const void *get_control() const { return m_control.data(); }
+    size_t get_controllen() const { return m_control.size(); }
+
+private:
+    tx_call_t m_opcode;
+    int m_flags;
+    std::vector<uint8_t> m_control;
+};
+
+using tx_call_ctx_ptr = std::shared_ptr<const tx_call_ctx>;
```
AlexanderGrissik: Do we really need shared_ptr? i mean, do we really need to refcount tx_call_ctx?

BasharRadya: one send call can generate multiple tx jobs.
i think it willl not be many tx jobs (most probebly 1,2) so i can simply converte it to deep copy without feeling the cost.

src/core/event/entity_context.cpp
```diff
@@ -161,8 +164,25 @@ void entity_context::add_incoming_socket(sockinfo *sock)
...
+             * that hold the encrypted record. Return them through reuse_buffer(), which
+             * unwraps them and releases the underlying buffers to the ring.
+             */
+            sockinfo_tcp *sock = static_cast<sockinfo_tcp *>(job.sock);
+            mem_buf_desc_t *buf = job.buf;
+
+            do {
+                mem_buf_desc_t *next = buf->p_next_desc;
+                buf->p_next_desc = nullptr;
+                sock->reuse_buffer(buf);
```
AlexanderGrissik: i would move all this handling to a seperate method, so not to trash this one.

src/core/event/entity_context.cpp
```diff
@@ -161,8 +164,25 @@ void entity_context::add_incoming_socket(sockinfo *sock)
...
     if (job.buf) {
-        /* coverity[check_return] */
-        job.buf->p_desc_owner->reclaim_recv_buffers(job.buf);
+        if (job.buf->lwip_pbuf.type == PBUF_ZEROCOPY &&
+            job.buf->lwip_pbuf.desc.attr == PBUF_DESC_TLS_RX) {
+            /* Zerocopy wrappers produced by the TLS receive path reference the ring buffers
+             * that hold the encrypted record. Return them through reuse_buffer(), which
+             * unwraps them and releases the underlying buffers to the ring.
+             */
+            sockinfo_tcp *sock = static_cast<sockinfo_tcp *>(job.sock);
```
AlexanderGrissik: guess it should be reinterpret_cast

src/core/event/entity_context.cpp
```diff
@@ -184,6 +204,20 @@ void entity_context::listen_socket_job(const job_desc &job)
     }
 }
 
+void entity_context::tls_setup_job(const job_desc &job)
+{
+#ifdef DEFINED_UTLS
+    sockinfo_tcp *sock = reinterpret_cast<sockinfo_tcp *>(job.sock);
+    assert(sock);
+    sockinfo_tcp_ops_tls *tls_ops = static_cast<sockinfo_tcp_ops_tls *>(sock->get_ops());
```
AlexanderGrissik: reinterpret_cast?

src/core/dev/hw_queue_tx.cpp
```diff
@@ -926,42 +936,56 @@ void hw_queue_tx::put_tls_dek(std::unique_ptr<dpcp::tls_dek> &&tls_dek_obj)
...
             return nullptr;
         }
-    } else {
-        tis.reset(m_tls_tis_cache.back());
-        m_tls_tis_cache.pop_back();
     }
 
     auto dek_obj = get_tls_dek(info->key, info->key_len);
     if (unlikely(!dek_obj)) {
+        m_tls_cache_lock.lock();
```
AlexanderGrissik: std guard?

src/core/dev/hw_queue_tx.cpp
```diff
@@ -926,42 +936,56 @@ void hw_queue_tx::put_tls_dek(std::unique_ptr<dpcp::tls_dek> &&tls_dek_obj)
         return;
     }
     // We don't allow unlimited DEK cache to avoid system DEK starvation.
-    if (likely(m_p_ring->tls_sync_dek_supported()) &&
-        m_tls_dek_put_cache.size() < safe_mce_sys().utls_high_wmark_dek_cache_size) {
-        m_tls_dek_put_cache.emplace_back(std::forward<std::unique_ptr<dpcp::tls_dek>>(tls_dek_obj));
+    if (likely(m_p_ring->tls_sync_dek_supported())) {
+        std::lock_guard<decltype(m_tls_cache_lock)> lock(m_tls_cache_lock);
```
AlexanderGrissik, line 940: why do we need the special cache lock? because it is accsed both from ring tx and rx locks? or because it is accessed outside the ring?

src/core/sock/sockinfo_tcp.h
```diff
@@ -384,7 +393,23 @@ class sockinfo_tcp : public sockinfo {
     tcp_timers_collection *get_tcp_timer_collection();
     bool is_cleaned() const { return m_is_cleaned; }
     static err_t rx_lwip_cb(void *arg, struct tcp_pcb *tpcb, struct pbuf *p, err_t err);
+    /*
+     * Deliver received data with the callback matching the socket ownership model.
+     * Keep this dispatch inline because the ownership mode is stable for the socket lifetime.
+     */
+    static inline err_t rx_lwip_cb_dispatch(void *arg, struct tcp_pcb *tpcb, struct pbuf *p,
```
AlexanderGrissik, line 400: Why is it needed? why not just rx_lwip_cb_entity_context to replace rx_lwip_cb if in threads mode?

src/core/sock/sockinfo_tcp.cpp
```diff
@@ -1412,7 +1438,8 @@ ssize_t sockinfo_tcp::tcp_tx_thread(xlio_tx_call_attr_t &tx_arg)
     return sent_bytes;
 }
 
-void sockinfo_tcp::tx_thread_commit(mem_buf_desc_t *buf, uint32_t offset, uint32_t size, int flags)
+void sockinfo_tcp::tx_thread_commit(mem_buf_desc_t *buf, uint32_t offset, uint32_t size, int flags,
```
AlexanderGrissik, line 1449: maybe this whole tx_thread_commit should be moved to tcp_ops and so you dont need to add that if to go to tls_ops. anyway we dont support udp but even if we did we could add udp_ops or somehing like this.

src/core/sock/sockinfo_tcp.cpp
```diff
@@ -1334,13 +1336,33 @@ ssize_t sockinfo_tcp::tcp_tx_thread(xlio_tx_call_attr_t &tx_arg)
         }
     }
 
+    if (tx_arg.opcode == TX_SENDMSG && tx_arg.attr.hdr &&
```
AlexanderGrissik: would move this preparation to seperate method.

src/core/sock/sockinfo_ulp.cpp
```diff
@@ -1593,7 +1808,8 @@ err_t sockinfo_tcp_ops_tls::recv(struct pbuf *p)
     tcp_recved(m_p_sock->get_pcb(), m_tls_rec_overhead, true);
     if (likely(pres)) {
         assert(pres->tot_len == (m_rx_rec_len - m_tls_rec_overhead));
-        err = sockinfo_tcp::rx_lwip_cb((void *)m_p_sock, m_p_sock->get_pcb(), pres, ERR_OK);
+        err =
+            sockinfo_tcp::rx_lwip_cb_dispatch((void *)m_p_sock, m_p_sock->get_pcb(), pres, ERR_OK);
```
AlexanderGrissik, line 1804: why not keep a pointer to the righ cb in the tls_ops. instead of branching each time

src/core/event/entity_context.h
BasharRadya: will remove shared ptr and do deep copy if needed

AlexanderGrissik: Wouldnt be better to just hold these few members of tx_call_ctx as plain members? how much copy overhead is this? the shared_ptr already has ptr and deleter 16 bytes.

AlexanderGrissik: i think with avx this whole struct can be copied at one shoot

src/core/sock/sockinfo_ulp.cpp
```diff
@@ -971,15 +1132,69 @@ ssize_t sockinfo_tcp_ops_tls::tx(xlio_tx_call_attr_t &tx_arg)
...
+            msg.msg_control = const_cast<void *>(tx_ctx->get_control());
+            msg.msg_controllen = tx_ctx->get_controllen();
+        }
+        tx_arg.opcode = TX_SENDMSG;
+        tx_arg.attr.flags |= tx_ctx->get_flags();
+        tx_arg.attr.hdr = &msg;
+    }
+
+    si_ulp_logdbg("Committing TLS TX job: store=%p offset=%u size=%u", store, offset, size);
+    tls_store_owner *owner = new (std::nothrow) tls_store_owner(store);
```
AlexanderGrissik, line 1159: why wouldnt the regular m_store release mechanism be enough?

src/core/sock/sockinfo.h
BasharRadya: will remove shared ptr and do deep copy if needed

src/core/sock/sockinfo_ulp.cpp
```diff
@@ -959,6 +1109,17 @@ ssize_t sockinfo_tcp_ops_tls::tx(xlio_tx_call_attr_t &tx_arg)
     }
 done:
 
+    if (worker_mode && m_next_recno_tx != last_recno) {
+        const uint32_t worker_record_overhead = static_cast<uint32_t>(
+            is_tx_tls13() ? TLS_13_RECORD_OVERHEAD : TLS_12_RECORD_OVERHEAD);
+        /*
+         * The app thread reserves only plaintext bytes, while ACK processing restores the full
```
AlexanderGrissik, line 1108: how it works in r2c mode?

src/core/sock/sockinfo_ulp.cpp
```diff
@@ -805,7 +928,37 @@ err_t sockinfo_tcp_ops_tls::tls_rx_consume_ready_packets()
...
+{
+    return tx_internal(tx_arg);
+}
+
+ssize_t sockinfo_tcp_ops_tls_thread::tx(xlio_tx_call_attr_t &tx_arg)
+{
+    return m_p_sock->tcp_tx_thread(tx_arg);
+}
+
+ssize_t sockinfo_tcp_ops_tls::tx_internal(xlio_tx_call_attr_t &tx_arg)
```
AlexanderGrissik, line 954: why not just to split this method to tx_internal_thread or have another one in sockinfo_tcp_ops_tls_thread? and avoid all the internal worker mode conditions?

src/core/sock/sockinfo_ulp.cpp
```diff
@@ -702,65 +764,82 @@ int sockinfo_tcp_ops_tls::setsockopt(int __level, int __optname, const void *__o
...
+            tcp_recv(m_p_sock->get_pcb(), sockinfo_tcp_ops_tls::rx_lwip_cb);
+            m_p_sock->unlock_tcp_con();
+
+            si_ulp_logdbg("Queueing TLS_SETUP job: direction=RX sock=%p", m_p_sock);
+            m_p_sock->get_entity_context()->add_job(
+                entity_context::job_desc {entity_context::JOB_TYPE_SOCK_TLS_SETUP,
+                                          entity_context::JOB_FLAG_TLS_RX, m_p_sock, nullptr, 0,
+                                          0});
+        } else {
+            m_p_sock->lock_tcp_con();
```
AlexanderGrissik: i would really refactor this whole method into context_setup _rx/tx_r2c/thread. but up to time.

src/core/sock/sockinfo_ulp.cpp
```diff
@@ -774,6 +853,50 @@ int sockinfo_tcp_ops_tls::setsockopt(int __level, int __optname, const void *__o
     return 0;
 }
 
+/*
+ * The TLS setup stage for worker thread mode. Runs on the owning entity context after
+ * setsockopt() reserved all HW resources on the application thread, so it cannot fail.
+ */
+void sockinfo_tcp_ops_tls::tls_setup_entity_context(int optname)
```
AlexanderGrissik, line 872: can we make the r2c path to reuse this method?

src/core/event/entity_context.cpp
BasharRadya: both works here..
wtih static_cast  u will have the C++ type safe/sematnic checks, while in reinterepret u will lose this check.

i see that the reinterepret is used in all other places so i will switch to reintrepret to keep up with the convention..

src/core/event/entity_context.cpp
BasharRadya: both works here..
wtih static_cast  u will have the C++ type safe/sematnic checks, while in reinterepret u will lose this check.

i see that the reinterepret is used in all other places so i will switch to reintrepret to keep up with the convention..

src/core/dev/hw_queue_tx.cpp
BasharRadya: Done

src/core/dev/hw_queue_tx.cpp
BasharRadya: Accessed from outside the ring, we don't want to hold the ring lock, so we use another dedicated lock.

src/core/sock/sockinfo_tcp.h
BasharRadya: pcb->recv already holds the TLS callback on a TLS socket, so it can't also hold the plaintext one — the mode choice has to happen where TLS delivers plaintext, and that code is shared by both modes.
i can add a new function pointer and call the pointer at the TLS delivering plaintext stage, but i don't see the performance gains with respect to the dispatch function (1 memory access at both, and the if condietion is neglable to the Branch prediector).

image

src/core/sock/sockinfo_tcp.cpp
BasharRadya: Done

src/core/sock/sockinfo_tcp.cpp
BasharRadya: @pasis  WDYT?
i can add new tcp_ops funciton

src/core/sock/sockinfo_ulp.cpp
BasharRadya: i can add a new function pointer and call the pointer at the TLS delivering plaintext stage, but i don't see the performance gains with respect to the dispatch function (1 memory access at both, and the if condition is neglable to the Branch predictor).

BasharRadya: but if you seom performance gain/ software design gain, i can do it, WDYT?

src/core/sock/sockinfo_ulp.cpp
BasharRadya: **m_store is a mem_buf_desc_t, not a mem_desc, so it cannot provide the get(), put(), and get_lkey() interface expected by **TLS zero-copy.** The wrapper adapts it to that existing interface**.

For TLS, m_store only backs the plaintext payload. One job may create several tls_records, and completions release those records, not m_store.

This follows the existing sendfile-ZC model:

sendfile: mapping_t (mem_desc) -> TLS records -> last record releases mapping
worker:   m_store -> tls_store_owner (mem_desc) -> TLS records -> last record releases m_store

src/core/sock/sockinfo_ulp.cpp
BasharRadya: In R2C, the full TLS record is charged automatically:

R2C sets XLIO_EXPRESS_MSG_SND_BUF, so tcp_tx_express() subtracts plaintext + TLS overhead.
ACK restores that same full size.

src/core/sock/sockinfo_ulp.cpp
BasharRadya: tx_internal() contains the shared TLS record state machine. Worker mode differs only in:

R2C-only send-credit polling.
Omitting XLIO_EXPRESS_MSG_SND_BUF.
Charging TLS overhead afterward.
Splitting it would duplicate record creation, IV/record-number updates, retries, rollback, and statistics. I preferred one shared implementation with three stable mode checks

src/core/sock/sockinfo_ulp.cpp
BasharRadya: Not directly.
but refactor to have the least duplicate code

BasharRadya: not directly.
refactored the code.

src/core/sock/sockinfo_ulp.cpp
BasharRadya: not directly.
refactored the code.

src/core/sock/sockinfo_tcp.h
AlexanderGrissik: Maybe then we should have tls_ops::rx_lwip_cb_entity_context?, common logic of undecrypting can go to helper function?

src/core/event/entity_context.cpp
AlexanderGrissik: Actually you are right. static_cast provides more compile time checks then reinterpret_cast.

src/core/sock/sockinfo_ulp.cpp
AlexanderGrissik: Because we dont want app thread to handle tls records and jsut push the plain bytes to the worker?

src/core/sock/sockinfo_ulp.cpp
AlexanderGrissik: The right design could be 
sockinfo_tcp_ops_tls_r2x : sockinfo_tcp_ops_tls
sockinfo_tcp_ops_tls_threads : sockinfo_tcp_ops_tls
:)
Maybe it could resolve other incompatabilities as well.
Anyway, at this stage the vtable should already be loaded so calling vfunc should be constant X work. Branching will depend on the predictor.

src/core/event/entity_context.cpp
AlexanderGrissik: Any comment here?

src/core/sock/sockinfo_ulp.cpp
BasharRadya: in worker thread, yea the app thread  push plain text, it does not know the context of kTLS (tls version etc..)

AlexanderGrissik (approved):

src/core/event/entity_context.cpp
BasharRadya: Done

PR #662, opened 2026-09-08
issue: 5245126 Keep socket properties on socket reuse

pasis:
## Description

socket_stats_t::reset() cleared the socket properties - socket type, offload state and the multicast fields - which are assigned only by the derived sockinfo constructors. sockinfo::socket_stats_init() is called not only from the base constructor, but also for a socket object that is reused in place for a new connection (syn_received_timewait_cb()), where no constructor runs. Every recycled socket therefore reported "???" instead of "TCP" and "Offloaded: No" for the rest of its life, and the tuning report attributed its traffic to the non-offloaded bucket.

Move the properties out of reset() into a new reset_connection() and use it on the re-initialization path. An object freshly taken from the sock_stats pool still gets the full reset, because the pool doesn't clean an object on release.

##### What
When a TIME_WAIT socket is reused, keep the socket type, offloading state and multicast interface.

##### Why ?
Fixes misleading information reported in statistics and tuning report when a TCP socket is reused.

##### How ?
architecture, approach, etc._

## Change type
- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TCP time-wait socket reuse by preserving stream socket behavior and offload status after socket statistics are reinitialized.

### DIFF
--- a/src/core/sock/sockinfo.cpp
+++ b/src/core/sock/sockinfo.cpp
@@ -156,17 +156,24 @@ sockinfo::~sockinfo()
 
 void sockinfo::socket_stats_init()
 {
-    if (!m_p_socket_stats) { // This check is for listen sockets.
+    if (!m_p_socket_stats) {
         m_p_socket_stats = sock_stats::instance().get_stats_obj();
         if (!m_p_socket_stats) {
             return;
         }
 
+        // The pool doesn't clean an object on release, so it still carries the
+        // properties of the previous socket. Drop them with the connection state.
+        m_p_socket_stats->reset();
+
         // Save stats as local copy and allow state publisher to copy from this location
         xlio_stats_instance_create_socket_block(m_p_socket_stats);
+    } else {
+        // Re-initialization of a reused socket object. Keep the properties: they are
+        // assigned by the derived constructor, which doesn't run on this path.
+        m_p_socket_stats->reset_connection();
     }
 
-    m_p_socket_stats->reset();
     m_p_socket_stats->fd = m_fd;
     m_p_socket_stats->inode = fd2inode(m_fd);
     m_p_socket_stats->b_blocking = m_b_blocking;
--- a/src/core/util/xlio_stats.h
+++ b/src/core/util/xlio_stats.h
@@ -253,14 +253,18 @@ struct socket_stats_t {
 #endif /* DEFINED_UTLS */
     socket_stats_t *_next_stat;
 
-    void reset()
+    /*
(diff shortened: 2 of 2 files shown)


src/core/sock/sockinfo.cpp
```diff
@@ -156,17 +156,24 @@ sockinfo::~sockinfo()
...
+        // The pool doesn't clean an object on release, so it still carries the
+        // properties of the previous socket. Drop them with the connection state.
+        m_p_socket_stats->reset();
+
         // Save stats as local copy and allow state publisher to copy from this location
         xlio_stats_instance_create_socket_block(m_p_socket_stats);
+    } else {
+        // Re-initialization of a reused socket object. Keep the properties: they are
+        // assigned by the derived constructor, which doesn't run on this path.
+        m_p_socket_stats->reset_connection();
```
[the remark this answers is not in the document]
pasis: Out of scope for the current PR.

src/core/sock/sockinfo.cpp
```diff
@@ -156,17 +156,24 @@ sockinfo::~sockinfo()
 
 void sockinfo::socket_stats_init()
 {
-    if (!m_p_socket_stats) { // This check is for listen sockets.
+    if (!m_p_socket_stats) {
```
tomerdbz, line 159: this is an oracle then to the Q: "is this a new socket or reuse" right?
BUT - what if the pool is exahsuted? you can mistakengly on recycle not call reset_connection but do a full reset

tomerdbz (changes requested):
I'd love a test here - can you add pls?

src/core/sock/sockinfo.cpp
pasis: you're right, if the pool is exhausted for the original socket and then has a free object on reuse, it doesn't matter what kind of reset we do - the information there cannot be consistent (because the original socket didn't set values, so there is nothing to keep). So, the design should be different.

pasis: Simplified the fix significantly to cover only possible TCP scenario. This removes corner cases and overcomplicated code for now. In the future we can revisit this area and make in more generic way with UDP support and so on.

Regarding the gtest, it becomes overcomplicated, because the required information is provided either by an external run-time interactive tool or in a file after a clean process exit. So I postpone it.

galnoam:
@tomerdbz, please review again.

